PR #5, opened 2015-07-17
initial RPC implementation

campagnola:
RPC implemented over zmq sockets.

* remote procedure calls can be synchronous, asynchronous, or no-return
* exceptions propagate nicely back to caller
* no specific event loop requirements
* client socket uses zmq.ROUTER, which allows us to manage multiple rpc servers from a single client socket.


samuelgarcia:
OK.
Need to implement the multiple client way later on.

PR #7, opened 2015-07-18
RPC updates

campagnola:
RPC now uses router-router zmq sockets.

* multiple clients can share a socket, as before
* a single server may have multiple clients
* servers now bind(), clients connect()
* ensure connection is established before sending procedure call requests
* more documentation and unit tests

### DIFF
--- a/pyacq/core/rpc.py
+++ b/pyacq/core/rpc.py
@@ -86,7 +86,7 @@ class RPCClientSocket(object):
         cmd = {'action': action, 'call_id': call_id,
                'args': args, 'kwds': kwds}
         cmd = json.dumps(cmd).encode()
-        #print("SEND:", name, cmd)
+        #print("SEND REQUEST:", self.socket.getsockopt(zmq.IDENTITY), name, cmd)
         self.socket.send_multipart([name, cmd])
         fut = Future(self, call_id)
         self.futures[call_id] = fut
@@ -94,6 +94,8 @@ class RPCClientSocket(object):
 
     def process(self):
         """Process all available incoming messages.
+        
+        Return immediately if no messages are available.
         """
         while True:
             try:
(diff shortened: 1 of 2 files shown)


pyacq/core/rpc.py
```diff
@@ -70,28 +94,56 @@ def send(self, name, action, *args, **kwds):
...
+            if timeout is None:
+                itimeout = -1
+            else:
+                dt = time.perf_counter() - start
+                itimeout = int((timeout - dt) * 1000)
+                if itimeout < 0:
+                    raise TimeoutError("Timeout waiting for Future result.")
+            try:
+                self.socket.setsockopt(zmq.RCVTIMEO, itimeout)
+                name = self.socket.recv()
```
campagnola: should use recv_multipart.

pyacq/core/rpc.py
```diff
@@ -101,45 +153,115 @@ def _process_msg(self, ident, msg):
...
+    ----------
+    name : str
+        The identifier used by the remote server.
+    addr : URL
+        Address of RPC server to connect to.
+    socket : None or RPCClientSocket
+        Optional RPCClientSocket object. Used to allow multiple RPCClients to
+        share a single zmq socket.
+    """
+    def __init__(self, name, addr, socket=None):
```
campagnola: `shared_client` or something instead of `socket`

```
client1 = RPCClient(...)
client2 = RPCClient(..., shared_client=client1)
```

pyacq/core/rpc.py
```diff
@@ -101,45 +153,115 @@ def _process_msg(self, ident, msg):
...
+        if socket is None:
+            socket = RPCClientSocket()
+        socket.connect(addr)
+        self._name = name.encode()
+        self._socket = socket
```
campagnola: `rpc_socket` instead of `socket`

PR #11, opened 2015-07-20
Manager

campagnola:
Manager unit tests are passing!

### DIFF
--- a/pyacq/core/host.py
+++ b/pyacq/core/host.py
@@ -1,4 +1,3 @@
-
 from .processspawner import ProcessSpawner
 from .rpc import RPCServer, RPCClient
 from .nodegroup import NodeGroup
@@ -22,12 +21,15 @@ class Host(RPCServer):
         self.nodegroup_process = {}
     
     def new_nodegroup(self, name, addr):
-        """
-        Create a new NodeGroup in a new process.
+        """Create a new NodeGroup in a new process.
+        
+        Return the RPC name and address of the new nodegroup.
         """
         assert name not in self.nodegroup_process, 'This node group already exists'
         #print(self._name, 'start_nodegroup', name)
-        self.nodegroup_process[name] = ProcessSpawner(NodeGroup, name, addr)
+        ps = ProcessSpawner(NodeGroup, name, addr)
+        self.nodegroup_process[name] = ps
+        return ps.name, ps.addr
     
     def close_nodegroup(self, name):
         """
@@ -40,4 +42,13 @@ class Host(RPCServer):
         client = RPCClient(name, self.nodegroup_process[name].addr)
         assert not client.any_node_running()
         self.nodegroup_process[name].stop()
+        del self.nodegroup_process[name]
 
+    def close_all_nodegroups(self):
+        """Close all NodeGroups belonging to this host.
+        """
+        for name, spawner in self.nodegroup_process:
+            spawner.kill()
+        self.nodegroup_process = {}
+        
+    
--- a/pyacq/core/manager.py
+++ b/pyacq/core/manager.py
@@ -13,11 +13,6 @@ class Manager(RPCServer):
         self.nodegroups = {}  # name:NodegroupProxy
         self.nodes = {}  # name:NodeProxy
         
-        # Over-simplified port management.
-        # Should probably have RPC servers automatically select port numbers
-        # instead.
-        self._next_port = 5000
-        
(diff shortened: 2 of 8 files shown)


pyacq/core/host.py
```diff
@@ -22,12 +21,15 @@ def __init__(self, name, addr):
...
+        """Create a new NodeGroup in a new process.
+        
+        Return the RPC name and address of the new nodegroup.
         """
         assert name not in self.nodegroup_process, 'This node group already exists'
         #print(self._name, 'start_nodegroup', name)
-        self.nodegroup_process[name] = ProcessSpawner(NodeGroup, name, addr)
+        ps = ProcessSpawner(NodeGroup, name, addr)
+        self.nodegroup_process[name] = ps
+        return ps.name, ps.addr
```
samuelgarcia: Why returning this ouputs ?

pyacq/core/manager.py
```diff
@@ -9,46 +9,102 @@ class Manager(RPCServer):
...
+        self.nodegroups = {}
+        self.rpc_hostname = addr.partition('//')[2].rpartition(':')[0]
 
-    def add_nodegroup(self, name):
-        return self.mgr.add_nodegroup(self.rpc_name, name)
+    def add_nodegroup(self, ng):
+        self.nodegroups[ng.rpc_name] = ng
+    
+    def list_nodegroups(self):
+        return list(self.nodegroups.keys())
```
samuelgarcia: By design this _Host class is too much, no ? This do redundently what pycaq.core.host.Host.
Could we have a automagic way ?
But we can leave it for the moement.

pyacq/core/manager.py
```diff
@@ -9,46 +9,102 @@ class Manager(RPCServer):
     """
     def __init__(self, name, addr):
         RPCServer.__init__(self, name, addr)
-        self._hosts = {}
```
samuelgarcia: I would have coose self._hostproxys = {}

pyacq/core/tests/test_host.py
```diff
@@ -5,15 +5,15 @@
...
-logging.basicConfig(level=logging.INFO)
+logging.getLogger().level=logging.INFO
 
 
 def test_host1():
     
-    process_host0  = ProcessSpawner(Host,  'host0', 'tcp://127.0.0.1:5406')
-    process_host1  = ProcessSpawner(Host,  'host1', 'tcp://127.0.0.1:5407')
+    process_host0  = ProcessSpawner(Host,  'host0', 'tcp://127.0.0.1:*')
+    process_host1  = ProcessSpawner(Host,  'host1', 'tcp://127.0.0.1:*')
```
samuelgarcia: In my mind, the Host port is fixed in advance, because a manager need to connect to Host.
It is more usefull for NodeGroup and in a case of only one Host on the same machine than the manager.

samuelgarcia:
I think I did not catch who have in charge of choosing the port.
I think it is good to let some object decide (Node, NodeGroup) if if we want to force on port it also good.
It is possible in this new version ?

pyacq/core/manager.py
campagnola: In this case they are not proxies anymore. I think we will only need the proxies in the context of the client / gui, so they are not so useful here. The classes below are only used as data structures for now (might get rid of them).

pyacq/core/manager.py
campagnola: I am not yet convinced that it is a good idea to re-use the same class to represent one object across multiple processes. `core.host.Host` is the *actual* Host, whereas this class is just a local representation of a host that the manager uses for bookkeeping.

pyacq/core/tests/test_host.py
campagnola: That's true, but for running unit tests we cannot be guaranteed access to a particular port.

campagnola:
You can still give a port number when you create any RPC server.

samuelgarcia:
OK. Can I merge ?

campagnola:
We now have many functions and classes that use `(name, addr)` in their signature. What do you think of combining these into a single argument like `name@tcp://host:port` ?

campagnola:
Yes, this is ok to merge.

samuelgarcia:
Not convince, of name@tcp://host:port.
In my mind the manager will a dict:
all_nodes[name] = addr, so better to not split the string any time.
Discussion is opened.

PR #14, opened 2015-07-21
Connection between two nodes.

samuelgarcia:


pyacq/core/node.py
```diff
@@ -32,11 +36,22 @@ def initialize(self, **kargs):
...
         raise(NotImplementedError)
     
-    def set_sources(self, ):
-        assert not self.isrunnng(), 'Cannot change source while running'
-        #TODO check the source list
-        #TODO create zmq.SUB socket
-        pass
+    def create_outputs(self, streamdef_list):
+        assert not self.running(), 'Cannot change source while running'
+        print('#####', self.name, len(self.out_streams))
```
campagnola: Remove or use `logging.info(...)` ?

pyacq/core/node.py
```diff
@@ -32,11 +36,22 @@ def initialize(self, **kargs):
...
+        assert len(self.out_streams)==0, 'Output Stream are already there'
+        # TODO check the compatibility are of the request and the Node possiobility
+        # todo check the len(self.out_streams) is the number of outputs
+        for streamdef in streamdef_list:
+            self.out_streams.append(StreamReceiver(**streamdef))
```
campagnola: Is this supposed to be streamsender?

samuelgarcia:
@campagnola:
The commit 8e21cfbbb45674febd00fe4e6cf543b635396406 can be remove if you dont like it.
It is just a try.

samuelgarcia:
Hi Luke:
this is a very long PR:
    * NodeGroup and QApplication more clean
    * Node and NodeWidget can be lanch locally or distantly
    * StreamSender and Stream receiver with socket plaindata or sharedarray
    * remotly register Node
    * some test with webcam

What I am not happy:
   * the way how Node connect them. We need some more deep discussion about that.
   * Connection bewteen local apps Node and distant nodes.

All is quite drafty but there tests.

example/simple_webcam.py
```diff
@@ -0,0 +1,34 @@
...
+# create a device in a new pocess
+man = create_manager()
+
+nodegroup = man.create_nodegroup()
+#dev = nodegroup.create_node('WebCamImageIO', name = 'cam0')
+dev = nodegroup.create_node('WebCamAV', name = 'cam0')
+
+dev.configure(camera_num = 0)
+stream_dict = dict(protocol = 'tcp', interface = '127.0.0.1', port = '9000',
+                    transfertmode = 'plaindata', streamtype = 'video',
```
campagnola: Should not need to specify stream type here--the node already knows (and enforces) this.

example/simple_webcam.py
```diff
@@ -0,0 +1,34 @@
...
+                    dtype = 'uint8', shape = (480, 640, 3), compression ='',
+                    scale = None, offset = None, units = '' , sampling_rate =30,
+                    )
+dev.create_outputs([ stream_dict ])    
+dev.initialize()
```
campagnola: Would rather see something like:

```
dev.output.configure(stream_spec)
viewer.input.connect(dev.output)
```

pyacq/core/nodegroup.py
```diff
@@ -56,38 +31,57 @@ class _NodeGroup(RPCServer):
...
+        self._readsocket_thread = ThreadReadSocket(self, parent = None)
+        self._readsocket_thread.new_message.connect(self._mainthread_process_one)
+        self._readsocket_thread.start()
+        self.app.exec_()
+    
+    def _mainthread_process_one(self, name, msg):
+        self._readsocket_thread.wait()
+        self._process_one(bytes(name), bytes(msg))
+        if self.running():
+            self._readsocket_thread.start()
```
campagnola: I imagine there is some overhead for starting / stopping the thread for every message. 
Maybe better to have the thread run in a loop and send it a message from the main thread when it needs to stop.

pyacq/core/nodelist.py
```diff
@@ -1,5 +1,8 @@
 # TODO find something fancy do resgister in another Host/Process
 
 all_nodes = { }
-def register_node(node_class):
-    all_nodes[node_class.__name__] = node_class
\ No newline at end of file
+def register_node(node_class, classname = None):
```
campagnola: Confusing terminology--register_node should maybe be register_node_type?

pyacq/core/rpc.py
```diff
@@ -12,13 +12,40 @@
...
 from logging import info
 
+import json
+try:
+    import msgpack
+    HAVE_MSGPACK = True
+except ImportError:
+    HAVE_MSGPACK = False
+
+class Serializer:
```
campagnola: should be two classes (JsonSerializer, MsgpackSerializer)

pyacq/core/rpc.py
```diff
@@ -240,7 +268,23 @@ def __init__(self, client, method):
         self.method = method
         
     def __call__(self, *args, **kwds):
-        return self.client._call_method(self.method, *args, **kwds)
+        # This little hack could be a help for manger remote method when dealing with node
+        # for instance:
+        # nodename = mynodeproxy.name
+        # manager.create_node_outputs(nodename, streamdef)
+        # could be directly be
+        # manager.create_node_outputs(mynodeproxy, streamdef)
```
campagnola: This is moving toward having proper proxy objects for all RPC interacion, which I think would be generally useful.
In pyqtgraph's multiprocess system, each remote call can either serialize and send back its return value, or it can keep a local reference to the return value and send back a new proxy object. This allows for a lot of flexibility in the way you control remote processes, particularly in that you can then feed those return values back into further remote calls without ever having transferred the data itself. For example:

```
# generate and process an array remotely without ever transferring it to the calling process
array_proxy = node_proxy.get_array()
node_proxy.process_data(array_proxy)
```

pyacq/core/stream.py
```diff
@@ -154,33 +262,55 @@ def __init__(self, **kargs):
         #send or cpy to buffer
         if self.params['transfermode'] == 'plaindata':
```
campagnola: Should probably break StreamSender and StreamReceiver up into subclasses (one class per transfer mode) rather than mix all code into monolithic classes?

pyacq/core/tests/test_sharedarray.py
```diff
@@ -0,0 +1,50 @@
...
+    np_a = sa.to_numpy()
+    np_a[:] = np.arange(10)
+    
+    sa2 = SharedArray(**sa.to_dict())
+    np_a2 = sa.to_numpy()
+    assert np_a is not np_a2
+    assert np.all(np_a ==np_a2)
+    
+
+def test_sharedarray():
```
campagnola: duplicate

pyacq/devices/npbufferdevice.py
```diff
@@ -0,0 +1,50 @@
+import numpy as np
+
+from ..core import Node, register_node
+from pyqtgraph.Qt import QtCore, QtGui
+
+
+class NumpyDeviceBuffer(Node):
```
campagnola: docstring would be helpful here.

pyacq/devices/npbufferdevice.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def stop(self):
+        self.timer.stop()
+        self._running = False
+    
+    def close(self):
+        pass
+    
+    def initialize(self):
+        assert len(self.out_streams)!=0, 'create_outputs must be call first'
```
campagnola: should assert len(outputstreams) == 1 ?

pyacq/devices/webcam_av.py
```diff
@@ -0,0 +1,83 @@
...
+        
+    def start(self):
+        print('/dev/video{}'.format(self.camera_num))
+        self.container = av.open('/dev/video{}'.format(self.camera_num), 'r','video4linux2')
+
+        self._thread = AVThread(self.out_streams[0], self.container)
+        self._thread.start()
+        self._running = True
+
+    def stop(self):
```
campagnola: There is a race condition here.  (same in webcam_imageio)

pyacq/devices/webcam_av.py
```diff
@@ -0,0 +1,83 @@
...
+    
+
+    def configure(self, camera_num = 0, **options):
+        self.camera_num = camera_num
+        self.options = options
+        #~ container = cv2.VideoCapture(camera_num)
+        #TODO deal with metadata
+        #~ del container
+    
+    def initialize(self):
```
campagnola: I guess this isn't finished?

pyacq/viewers/imageviewer.py
```diff
@@ -0,0 +1,63 @@
+from ..core import WidgetNode
+from pyqtgraph.Qt import QtCore, QtGui
+
+import numpy as np
+import vispy
```
campagnola: Let's use pyqtgraph for now; vispy's images still need some work (I plan to start on this soon)

pyacq/core/tests/test_node.py
```diff
@@ -0,0 +1,144 @@
...
+    receiver0.configure()
+    receiver0.set_inputs([ streamdef0 ])
+    receiver0.initialize()
+    receiver0.show()
+    
+    
+    #receiver1 is in local QApp
+    from pyqtgraph.Qt import QtCore, QtGui
+    from pyacq.core.tests.fakenodes import ReceiverWidget
+    app = QtGui.QApplication([])
```
campagnola: This will crash if any other unit tests need to use a qapplication.

example/simple_webcam.py
samuelgarcia: Yes. And the stream type is not really clear now.

example/simple_webcam.py
samuelgarcia: I am agree. We will need a  proxy object for this outputs and inputs.

pyacq/core/nodegroup.py
samuelgarcia: Yes. Maybe there is something more fancy.
But for synchronous reason we need to deal with one message before getting the next one.
Maybe a queue in between the thread and the app + signals.
We need to process the message in the main App.

pyacq/core/nodelist.py
samuelgarcia: Yes.

pyacq/core/rpc.py
samuelgarcia: OK.

pyacq/core/rpc.py
samuelgarcia: OK.
I remove this for the moment.
And we will think about a global solution in a separate PR.

pyacq/core/stream.py
samuelgarcia: Maybe for implementation.
But I like the idea to deal with StreamSender or StreamReceiver inside nodes without worring wich transfer mode they are, so one class.

pyacq/core/tests/test_node.py
samuelgarcia: Why ?

pyacq/core/tests/test_sharedarray.py
samuelgarcia: oups.

pyacq/devices/npbufferdevice.py
samuelgarcia: OK.

pyacq/devices/npbufferdevice.py
samuelgarcia: Yes.

pyacq/devices/webcam_av.py
samuelgarcia: Sur, it is completly drafty.

pyacq/devices/webcam_av.py
samuelgarcia: I don't get your comment.

pyacq/viewers/imageviewer.py
samuelgarcia: OK.

samuelgarcia:
Hi luke I have push some changes and try to fix your conerns.
Tell me your thnogth.

pyacq/core/tests/test_node.py
campagnola: You can only create one QApplication per process, even if an older QApplication has been exited. On some systems, creating another simply causes the program to lock up.

pyacq/devices/webcam_av.py
```diff
@@ -0,0 +1,83 @@
...
+    
+    def start(self):
+        self.container = av.open('/dev/video{}'.format(self.camera_num), 'r','video4linux2', self.options)
+
+        self._thread = AVThread(self.output, self.container)
+        self._thread.start()
+        self._running = True
+
+    def stop(self):
+        self._thread.running = False
```
campagnola: Race condition: What happens if both threads try to access this `running` attribute at the same time? You can end up with crashes. Every variable that is to be touched by multiple threads must be protected by a mutex.

samuelgarcia: Normally python do this for us. This is the role of GIL. No ?
AVThread.running is read by the thread and write by the node.
It should cause problem. No ?
Do you have something a crash scenario in mind ?

pyacq/core/tests/test_node.py
samuelgarcia: I don't get the point.
You can't create a new QApplication for each test ?
We need a big one for tests ?
Or is it only at module level ?

campagnola: That's correct--the unit tests should create only one QApplication (per process) and all tests within that process must share it.

samuelgarcia: on which system is buggy ?
Onlinux and windows I don't have this.
This is not convenient at all.
If you write one test, you need to read the entire module to see if exist another QApp ?

pyacq/devices/webcam_av.py
campagnola: In general the GIL does not protect you from race conditions. Even though only one thread can run the python interpreter at a time, it is possible to get a crash if the interpreter locks in the middle of an operation in one thread, and then unlocks with a conflicting operation in another thread. 

In this case you _might_ be safe because the operation is so simple that it might be atomic from the standpoint of the GIL, but you are much better safe than sorry in cases like this, since the cost of being wrong is so high (in two years, when the code has become very complex, you might notice some random segmentation faults and wonder what cases them). I've been through this many times, and I have wasted weeks of time fixing these issues.

My personal rules to avoid threading-related crashes are: 
1. All threads are implemented as a class
2. Only the thread instance may access its own attributes; everyone else must go through a method
3. These methods must handle proper locking of attributes

pyacq/core/tests/test_node.py
campagnola: Just use something like:
```
app = QApplication.instance()
if app is None:
    app = QApplication([])
```
This is what `pyqtgraph.mkQApp()` does.

pyacq/devices/webcam_av.py
samuelgarcia: point 3 lead to a mutex for self.running() and then you have have a mutux at each loop ?

campagnola: The only attribute that is accessed from both threads is `running`. So you would need something like:
```
class Thread(QtCore.QThread):
    def __init__(self):
        self.lock = QtCore.QMutex()
        self.running = False

    def run(self):
        with self.lock:
            self.running = True
        while True:
            with self.lock:
                if not self.running:
                    break

    def stop(self):
        with self.lock:
            self.running = False
```
It's a lot of extra work, but get used to it--even it it turns out to be unnecessary in this case, you're much better off not trying to guess which cases might be unsafe and being consistent about thread handling.

samuelgarcia:
Travis is complaining about the size.
"Estimated content size difference = 148 kB"
I guess it is in the .travis.yaml.

What shall we do ?

pyacq/devices/webcam_av.py
samuelgarcia: OK.

campagnola:
You can ignore that error--it's just a warning to make sure we don't
accidentally merge large files.

pyacq/devices/webcam_av.py
samuelgarcia: Which version of pyqt support the QMutex.__enter__ and QMutex.__exit__ for the with ?

campagnola: Ummm.. haha, no version. 
It is here: https://github.com/pyqtgraph/pyqtgraph/blob/develop/pyqtgraph/util/mutex.py
You could just copy `mutex.py` if you don't want pg to be a dependency?
Or you can use `threading.Lock`, but pg's mutex provides deadlock debugging, which is bound to come in handy sooner or later..

samuelgarcia: No problem for dependency.
I am a pyqtgraph lover and I trust you.

samuelgarcia:
Any other comments for this very long PR ?

example/simple_webcam.py
```diff
@@ -0,0 +1,27 @@
...
+dev.configure(camera_num = 0)
+dev.output.configure(protocol = 'tcp', interface = '127.0.0.1', transfertmode = 'plaindata')
+dev.initialize()
+
+#view is a Node in local QApp
+app = QtGui.QApplication([])
+
+viewer = ImageViewer()
+viewer.configure()
+viewer.input.connect(dev.output)
```
campagnola: Yay!

samuelgarcia: There is still 2 problems in that design:
  * the manager do not know this connection
  * the node do not known the node.name and output.name whit wich it is connected. (easy to fix).

Someting to add in a new PR.

example/simple_webcam.py
```diff
@@ -0,0 +1,27 @@
...
+from pyqtgraph.Qt import QtCore, QtGui
+
+
+man = create_manager()
+
+# this create the dev in a separate process (NodeGroup)
+nodegroup = man.create_nodegroup()
+dev = nodegroup.create_node('WebCamAV', name = 'cam0')
+dev.configure(camera_num = 0)
+dev.output.configure(protocol = 'tcp', interface = '127.0.0.1', transfertmode = 'plaindata')
```
campagnola: This is much nicer too!

samuelgarcia:
Do you anderstand the travis segmentation fault ?
It occurs after all tests pass ?
It is in rpc stuff...
Any ideas ?

pyacq/core/nodegroup.py
```diff
@@ -1,50 +1,31 @@
...
-        self.rpc_thread = RpcThread(self.rpc_server, parent = None)
-        self.rpc_thread.finished.connect(self.app.quit)
-        self.rpc_thread.start()
-        self.app.exec_()
     
+    def run(self):
+        self.go = True
+        while self.go:
+            # we need to poll in case of RPCServer.close()
+            event = self.rpc_server._socket.poll(500)
```
campagnola: I think you have two threads accessing the same socket? This is a good read:
http://zguide.zeromq.org/page%3aall#Multithreading-with-ZeroMQ

campagnola:
No clue about the segfault. This is exactly why it's so important to be meticulous about threading--in my experience with acq4, there are two things that cause segfaults more than any other: multithreading mistakes (which is why the 0MQ folks consider "shared state" to be a fundamentally broken programming model), and misuse of Qt (which is why I started this: http://stackoverflow.com/questions/11945183/what-are-good-practices-for-avoiding-crashes-hangs-in-pyqt ).

I would start by adding `faulthandler.enable(all_threads=True)` to see if we can get more information about the crash. If you can make it crash locally, that will greatly improve the chances of discovering the cause. If it crashes when the application exits, then it's very likely to be caused by some mishandling of Qt (perhaps creating QObjects without parents).

pyacq/core/nodegroup.py
samuelgarcia: I do not see it.

Only the ThreadReadSocket poll and then recv the message.
And the the message is transimeted by QSignal to the main thread.
Could you explain ? please.

I have read somewhere that there is no problem in zmq creating the socket in one thread and using it other thread.

samuelgarcia:
I don't think that it is QtThread relative. It is in RPC and so inanother process.
The strange thing is that it occurs AFTER that all tests are passed!!
How is it possible ?

campagnola:
If it's after the tests have passed, then it is probably caused by Qt tearing down incorrectly when python exits (which is usually fixed by making sure the QObject hierarchy is set up correctly). I'm trying to run the tests on my end but I seem to have a broken pytest..

pyacq/core/nodegroup.py
campagnola: `ThreadReadSocket.run()` is reading from the socket, and `NodeGroup._process_one()` is sending on the socket.

From the link: "This may appear to work at first but will fail randomly in real use. Remember: Do not use or close sockets except in the thread that created them."

samuelgarcia: Oups. I totaly missed the fact that _process_one() also return via the socket.

I see 2 solutions:
  1 - get back to older solution (start/stop thread at each message)
  2 - Decompose in 3 steps : a) Thread socket.recv and emit it with QSIgnal b) main apply the fn() and emit the ret with QSignal c) Thread socket._send_result

I prefer the first one but I don't known the impact of start()/stop() on perf.

campagnola: I think (1) had exactly the same problem--wasn't it reading in a thread and writing from main? 
Other options:
* Don't use a thread at all; just poll with QTimer
* Put all of NodeGroup + RPC in a worker thread, and leave the GUI in the main thread. Do the RPC commands really need to interact with the GUI thread at all?

samuelgarcia: For (3) :I do not like so much QTImer. because message are not executed immediatly but every XXms.
It is less responssive.
For (4): we need the NodeGroup to be in the main thread because NodeGroup.nodes['mynode'].method1() need to be call safely. A Node can be a widget. I made some test on window and linux and it works.

For the solution is :+1: 
   * recv and send in a thread or good mutex.
   * fn() in the main

No solution (1) was:
   - a thread is wating in background on recv (no poll).
   - when a message come the thread emit a message and stop.
   - the main thread process the message : exec fn() + send back result
   - the main thread start the thread again.
It was not so bad but start/stop overhead.

samuelgarcia: I made a small and naive benchmark. start/stop a thread really afect speed. So (1) is out.

I go for solution (2). Tell me.

samuelgarcia: I finnaly came back to solution 1.
Because was a nightmare  for perf.
Yes, starting/stopping take time (only 10000 start/stop cycles on my machine) but it is small compare to the RPC machinery.

I have made sa bench code in test_nodegroup solution 1 is not so bad:

Host  : sync ping per second 1895.0839374586299
Host  : async ping per second 10865.453264321723
NodeGroup  : sync ping per second 1421.3407280166482
NodeGroup  : async ping per second 3402.8876203580803

campagnola: > a thread is wating in background on recv (no poll).
> when a message come the thread emit a message and stop.
> the main thread process the message : exec fn() + send back result
> the main thread start the thread again. It was not so bad but start/stop
overhead.

Same problem: the background thread reads from the socket, whereas the main
thread is sending on the socket. This is not allowed.

samuelgarcia: I don't follow your logic : I want Node can be widget.
For the end user it is so simple. You can create local node/widget and remove node/widget and minupulated them the same way.
And it already works.
have a deep look in test_node and test_nodegroup
the overhead is small. no ? speed ratio is about 1.3 and 3

samuelgarcia: I have added example/remote_or_local.py

It show a behavior I like. A node is always a Node local or remote. And a Node can be Widget. remote or local.

If you have in mind another implementation I would like this continuing to work.

campagnola: > I don't follow your logic : I want Node can be widget.

Why is it necessary for nodes to be widgets? It make more sense to me
if nodes can create/control widgets from a separate thread.

> For the end user it is so simple. You can create local node/widget and remove node/widget and minupulated them the same way.

I am not suggesting any API changes; the end user won't notice the difference.

> And it already works.

I strongly recommend to heed the "may cause random crashes later on" warning.

campagnola: I like the `local_or_remote` example, and I see why you want nodes to be
Widgets. We can make this happen, but it will require some extra effort to
establish communication between the rpc and GUI threads. Basically we need
something like a Qt RPC that works by passing signals one way to make
requests, and passing zmq messages to send the return values back. I think
this can be more or less transparently built into the existing RPC system.

samuelgarcia: Cool if you like.
Note that it already works. Actual NodeGroup is more or less th Qt RPC. No ?

campagnola: > Note that it already works.

I'll quote the zmq manual again:  "This may appear to work at first but
will fail randomly in real use. Remember: Do not use or close sockets
except in the thread that created them."

This point cannot be overstated.

pyacq/core/nodegroup.py
```diff
@@ -1,93 +1,161 @@
...
-    def __init__(self, rpc_server, parent = None):
+class RpcThreadSocket( QtCore.QThread):
+    new_message = QtCore.Signal(QtCore.QByteArray, QtCore.QByteArray)
+    def __init__(self, rpc_server, local_addr, parent = None):
         QtCore.QThread.__init__(self, parent)
-        self.rpc_server = rpc_server
+        self.rpc_socket = weakref.ref(rpc_server._socket)
+        
+        context = zmq.Context.instance()
+        self.local_socket = context.socket(zmq.PAIR)
```
campagnola: This class looks great!
One minor issue: the sockets and poller should be created in the thread where they are used.

samuelgarcia: You say this for local_socket or rpc_socket ?
I have read in zmq that it is sane to instanciate socket in a thread and use them in another thread when there a guarantee that the socket is not manipulated in the 2 places.
In http://zguide.zeromq.org/page:all, they said
"Technically it's possible to migrate a socket from one thread to another but it demands skill."

campagnola: This is skill that neither you nor I have. Isn't it easy enough to just move the socket creation into the thread?

samuelgarcia: Yes you have this skill. I admit that I am border line.
I already merge. Sorry.
I will do a new PR for this (and other small thing).

local_socket is easy to move in run. But for rpc_socket, it is big piece of work, because socket is created at RPCServe instanciation due to rpc.py and processspawner.py.

campagnola: Ok, if it is a lot of work then we'll deal with it later.

PR #18, opened 2015-08-02
travis CI

campagnola:
Importing `.travis.yml` from pyqtgraph. This includes some extra features:

* Check that no files have CRLF
* Check total repository size does not grow too much
* Test install
* Nice colored output

### DIFF
--- a/pyacq/core/stream.py
+++ b/pyacq/core/stream.py
@@ -70,8 +70,9 @@ class StreamSender:
         context = zmq.Context.instance()
         self.socket = context.socket(zmq.PUB)
         self.socket.bind(self.url)
+        self.addr = self.socket.getsockopt(zmq.LAST_ENDPOINT).decode()
+        self.port = self.addr.rpartition(':')[2]
         
-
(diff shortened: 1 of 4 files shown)


pyacq/core/stream.py
```diff
@@ -134,8 +134,6 @@ def _compress_blosclz4(self, index, data):
         return index, data
     
     def close(self):
-        if self.params['protocol'] == 'tcp':
-            self.socket.unbind(self.url)
```
campagnola: I had to remove these to avoid a zmq bug:
https://github.com/zeromq/libzmq/issues/949

Any reason you think it is needed? I presume calling `socket.close()` will unbind it as well.

campagnola:
All tests passing!

I ran into a weird issue with the stream tests--they would lock up at a different random location every time the tests were run. I fixed this by having travis install pyzmq from pip instead of conda (to get the most recent version).

samuelgarcia:
Cool.
Thanks.

PR #19, opened 2015-09-04
Device pyaudio

samuelgarcia:
Simple PyAudio Node : grab and playback.

### DIFF
--- a/pyacq/core/tests/test_tools.py
+++ b/pyacq/core/tests/test_tools.py
@@ -1,0 +1,64 @@
+from pyacq.core.stream  import OutputStream, InputStream
+from pyacq.core.tools  import ThreadPollInput
+from pyqtgraph.Qt import QtCore, QtGui
+import pyqtgraph as pg
+
+import numpy as np
+import weakref
+import time
+
+
+def test_ThreadPollInput():
+    app = pg.mkQApp()
+    
+    nb_channel = 16
+    chunksize = 1024
+    stream_spec = dict(protocol = 'tcp', interface = '127.0.0.1', port='*', 
+                       transfermode = 'plaindata', streamtype = 'analogsignal',
+                       dtype = 'float32', shape = (-1, nb_channel), compression ='',
+                       scale = None, offset = None, units = '')
+    
+    class ThreadSender(QtCore.QThread):
+        def __init__(self, output_stream, parent = None):
+            QtCore.QThread.__init__(self, parent)
+            self.output_stream = weakref.ref(output_stream)
+        
+        def run(self):
+            index = 0
+            for i in range(5):
+                index += chunksize
+                arr = np.random.rand(chunksize, nb_channel).astype(stream_spec['dtype'])
+                self.output_stream().send(index, arr)
+                time.sleep(0.05)
+            self.terminated.emit()
+    
+    
+    outstream = OutputStream()
+    outstream.configure(**stream_spec)
+    instream = InputStream()
+    instream.connect(outstream)
+    
+    sender = ThreadSender(output_stream = outstream)
+    poller = ThreadPollInput(input_stream = instream)
+    
+    def on_new_data(pos, arr):
+        print(pos, arr.shape)
+    
+    def terminate():
+        sender.wait()
+        poller.stop()
+        poller.wait()
+        app.quit()
+    
+    sender.terminated.connect(terminate)
+    poller.new_data.connect(on_new_data)
+    
+    poller.start()
+    sender.start()
+    
+    app.exec_()
+
+    
+
+if __name__ == '__main__':
+    test_ThreadPollInput()
--- a/pyacq/core/tests/test_tools.py
+++ b/pyacq/core/tests/test_tools.py
@@ -16,7 +16,7 @@ def test_ThreadPollInput():
     stream_spec = dict(protocol = 'tcp', interface = '127.0.0.1', port='*', 
                        transfermode = 'plaindata', streamtype = 'analogsignal',
                        dtype = 'float32', shape = (-1, nb_channel), compression ='',
-                       scale = None, offset = None, units = '')
+                       scale = None, offset = None, units = '', timeaxis = 0)
     
     class ThreadSender(QtCore.QThread):
         def __init__(self, output_stream, parent = None):
@@ -41,8 +41,14 @@ def test_ThreadPollInput():
     sender = ThreadSender(output_stream = outstream)
     poller = ThreadPollInput(input_stream = instream)
     
+    
+    global last_pos
+    last_pos= 0
     def on_new_data(pos, arr):
-        print(pos, arr.shape)
+        assert arr.shape==(chunksize, nb_channel)
+        global last_pos
+        last_pos += chunksize
+        assert last_pos==pos
     
     def terminate():
         sender.wait()
--- a/pyacq/core/tools.py
+++ b/pyacq/core/tools.py
@@ -1,0 +1,38 @@
+from pyqtgraph.Qt import QtCore, QtGui
+from pyqtgraph.util.mutex import Mutex
+import weakref
+import numpy as np
+
+
+class ThreadPollInput(QtCore.QThread):
+    """
+    Thread that pool in backgroup an InputStream (zmq.SUB).
+    And emit Signal.
+    Util for Node that have inputs.    
+    """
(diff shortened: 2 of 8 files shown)


pyacq/core/tools.py
```diff
@@ -0,0 +1,38 @@
...
+    def run(self):
+        with self.lock:
+            self.running = True
+        
+        while True:
+            with self.lock:
+                if not self.running:
+                    break
+            
+            ev = self.input_stream().poll(timeout = self.timeout)
```
campagnola: Here it is again..  Perhaps I'll spend some time next month cleaning these up.

pyacq/core/tests/test_tools.py
```diff
@@ -0,0 +1,64 @@
...
+    outstream = OutputStream()
+    outstream.configure(**stream_spec)
+    instream = InputStream()
+    instream.connect(outstream)
+    
+    sender = ThreadSender(output_stream = outstream)
+    poller = ThreadPollInput(input_stream = instream)
+    
+    def on_new_data(pos, arr):
+        print(pos, arr.shape)
```
campagnola: Would be nice to verify that the data is received correctly rather than printing.

pyacq/devices/audio_pyaudio.py
```diff
@@ -0,0 +1,156 @@
...
+    def _close(self):
+        self.audiostream.close()
+        self.pa.terminate()
+        del self.thread
+
+    def _audiocallback(self, in_data, frame_count, time_info, status):
+        #~ print('audiocallback', len(self.out_queue))
+        if in_data is not None:
+            self.head += self.chunksize
+            self.output.send(self.head, in_data)
```
campagnola: ..and this is another. 
However: I am reading about how to use sockets safely from multiple threads, and some people say it is easier than the documentation suggests. Might turn out not to be an issue!

pyacq/devices/audio_pyaudio.py
```diff
@@ -0,0 +1,156 @@
...
+
+    def _close(self):
+        self.audiostream.close()
+        self.pa.terminate()
+        del self.thread
+
+    def _audiocallback(self, in_data, frame_count, time_info, status):
+        #~ print('audiocallback', len(self.out_queue))
+        if in_data is not None:
+            self.head += self.chunksize
```
campagnola: To be safe, I would rename `head` to `_head`, and I would make `chunksize` into a read-only property.

campagnola:
Code looks good to me. Haven't been table to get the test running, though: 

```
  File "pyacq/devices/tests/test_audio_pyaudio.py", line 76, in <module>
    test_local_app_in_out()
  File "pyacq/devices/tests/test_audio_pyaudio.py", line 26, in test_local_app_in_out
    format = 'int16', chunksize = 1024)
  File "/home/luke/work/pyacq/pyacq/core/node.py", line 111, in configure
    self._configure(**kargs)
  File "/home/luke/work/pyacq/pyacq/devices/audio_pyaudio.py", line 75, in _configure
    rate=int(self.sampling_rate), input_device=self.input_device_index),\
  File "/home/luke/work/pyacq/venv3/lib/python3.4/site-packages/pyaudio.py", line 934, in is_format_supported
    return pa.is_format_supported(rate, **kwargs)
ValueError: ('Invalid number of channels', -9998)
```

I'll add this to my to-fix list.

pyacq/devices/audio_pyaudio.py
samuelgarcia: I think there is a field in between your suspiciousness thread approach and my over confidence thread approach.

pyacq/devices/audio_pyaudio.py
samuelgarcia: OK. chunlsize can be set at when Node.configure()

samuelgarcia:
This error come from pyaudio when you chosse a pair of input/ouput whihch is not full duplex.

Try others input_device_index/output_device_index.

Playing while grabbing is not so easy in pyaudio it depends on hostapi.

campagnola:
Aha! I probably don't have full duplex on any of my built-in devices.

campagnola:
Grumbly segfaulty. (I'm back!)

samuelgarcia:
I spend lot of time on this. Any help would be appreciate.
With PyQT5 , there is no segfault.

the main problem is that in the same test module I app.exec_() and app.quit() several times for different test. And it is hard to known when object are garbage collected and remove by qt4 itself. I guesss you have more experiment than me. I never before start/stop the same app several app.

Note that if there is only one test then no seg fault.

campagnola:
I culled it down to two tests required to cause the fault:

```
pyacq/test_segfault.py::test_visual_node_both_in_main_qapp_and_remote_qapp PASSED
pyacq/test_segfault.py::test_ThreadPollInput PASSED

===================================================================================================== 2 passed in 6.91 seconds ======================================================================================================
Segmentation fault
```

Removing `app.exec_() / quit()` from the first test did not help.

pyacq/core/tests/test_tools.py
```diff
@@ -0,0 +1,70 @@
...
+    chunksize = 1024
+    stream_spec = dict(protocol = 'tcp', interface = '127.0.0.1', port='*', 
+                       transfermode = 'plaindata', streamtype = 'analogsignal',
+                       dtype = 'float32', shape = (-1, nb_channel), compression ='',
+                       scale = None, offset = None, units = '', timeaxis = 0)
+    
+    class ThreadSender(QtCore.QThread):
+        def __init__(self, output_stream, parent = None):
+            QtCore.QThread.__init__(self, parent)
+            self.output_stream = weakref.ref(output_stream)
```
campagnola: Why use a weakref here?

pyacq/core/tools.py
```diff
@@ -0,0 +1,38 @@
...
+class ThreadPollInput(QtCore.QThread):
+    """
+    Thread that pool in backgroup an InputStream (zmq.SUB).
+    And emit Signal.
+    Util for Node that have inputs.    
+    """
+    new_data = QtCore.Signal(int,object)
+    def __init__(self, input_stream, timeout = 200, parent = None):
+        QtCore.QThread.__init__(self, parent)
+        self.input_stream = weakref.ref(input_stream)
```
campagnola: ..and why a weakref here?

pyacq/core/tests/test_tools.py
samuelgarcia: To avoid that a QThread have a reference to the Stream that belong to the Node.
This avoid loop reference.
Node.thread
Node.input
thread.input << weakref here.

Furthermore Stream IS a zmq socket and if the Node disapear (with its input_stream) then no worry in the thread.

Are you agree ?

pyacq/core/tools.py
samuelgarcia: Same reason.

pyacq/core/tests/test_tools.py
campagnola: I think the zmq socket should belong to the thread that is reading it. Ideally the socket would be created in the same thread, but since that is sometimes difficult, a compromise would be to create the socket in one thread, pass it to the poller thread, and then delete the reference from the original thread, to avoid any possibility of touching the socket from the wrong thread.

For example: zmq sockets probably do some cleanup when they are collected. If this cleanup happens in one thread while the socket is being polled in another, then you might crash.

samuelgarcia: No it is not possible in the actual design. See Node docstring.
Node.outputs[''].configure and Node.intputs[''].connect are done in between Node.configure() and Node.initialize().
So InputStream and OutPutStream belong to the Node. All node do not necessary have a thread.
Node.outputs[''].configure imply a socket.bind and Node.intputs[''].connect imply a socket.connect.
Threads are created at Node.initialize steps if necessary.

node.doc:
Usage of a Node, the **order** is important:
        1 - Instanciate : directly or remotly Node with NodeGroup.creat_node
        2 - Node.configure(...)
        3 - Connect Input (if have input) : Node.inputs['input_name'].connect(other_node.outpouts['output_name'])
        4 - Configure outputs : Node.outputs['output_name'].configure(...)
        5 - initialize : Node.initialize(), this also check inputs and outputs specifications.
        6 - Start/Stop many time : Node.start()/Node.stop()
        7 - Node.close() : done by NodeGroup if the Node is remoted

PR #21, opened 2015-09-16
Viewer

samuelgarcia:
Starting to port old oscilloscope viewer (based on pyqtgraph).
This PR also include #19 and #20. Please merge after.

### DIFF
--- a/pyacq/core/__init__.py
+++ b/pyacq/core/__init__.py
@@ -7,4 +7,4 @@ from .nodelist import register_node_type
 from .manager import Manager, create_manager
 from .stream import OutputStream, InputStream
 from .sharedarray import SharedArray
-from .tools import ThreadPollInput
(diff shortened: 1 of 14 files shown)


campagnola:
Can't figure out why this says there are conflicts--it merges without problems for me. I'm just going to manually merge/push.

samuelgarcia:
Are you happy with this oscilloscope ?
I will add some changes in a near futur too. Because I want to add ChannelProperties object (compatible with RPC) than handle name/index/units/color for channels than can be share with multiple viewers. (ex: 2 oscillosope one filtered the other not)

campagnola:
It is good--it certainly works well as a starting point, and I guess it is pretty consistent with what you have used in the past. I will probably be making suggestions as we get into more complex use cases.

PR #24, opened 2015-10-02
Viewer tfr

samuelgarcia:
First step is done. It is quite efficient. But stil have some crash chen local workers.
This QTimerFreq can be improvred when RPC will be improved too.
This depend on PR #21.

### DIFF
--- a/pyacq/core/client.py
+++ b/pyacq/core/client.py
@@ -12,7 +12,7 @@ class ManagerProxy(RPCClient):
     def connect_host(self, name, addr):
         self._call_method('connect_host', name, addr)
         if name not in self._host_proxy:
-            self._host_proxy[name] = HostProxy(name, addr)
+            self._host_proxy[name] = HostProxy(self, name, addr)
         return self._host_proxy[name]
     
     def default_host(self):
@@ -122,14 +122,10 @@ class InputStreamProxy:
     
     def connect(self, output):
         if isinstance(output, dict):
-            print('InputStreamProxy.connect', output)
             self.node.connect_input(self.name, output)
         else:
-            print('InputStreamProxy.connect', output.params)
             self.node.connect_input(self.name, output.params)
-        
-        
-
+    
     @property
     def params(self):
         return self.node.get_output(self.name)
--- a/pyacq/core/client.py
+++ b/pyacq/core/client.py
@@ -78,12 +78,19 @@ class NodeProxy(object):
     def __init__(self, nodegroup, name):
         self.nodegroup = nodegroup
         self.name = name
-        self.inputs = { name:InputStreamProxy(self, name) for name in self.nodegroup.get_node_attr(self.name, '_input_specs').keys()}
-        self.outputs = { name:OutputStreamProxy(self, name) for name in self.nodegroup.get_node_attr(self.name, '_output_specs').keys()}
-        
+        #~ self.inputs = { name:InputStreamProxy(self, name) for name in self.nodegroup.get_node_attr(self.name, '_input_specs').keys()}
+        #~ self.outputs = { name:OutputStreamProxy(self, name) for name in self.nodegroup.get_node_attr(self.name, '_output_specs').keys()}
         
     def __getattr__(self, name):
         return lambda *args, **kwargs: getattr(self.nodegroup, 'control_node')(self.name, name, *args, **kwargs)
+
+    @property
+    def inputs(self):
+        return { name:InputStreamProxy(self, name) for name in self.get_input_names() }
+
+    @property
+    def outputs(self):
+        return { name:OutputStreamProxy(self, name) for name in self.get_output_names() }
 
     @property
     def input(self):
@@ -115,8 +122,10 @@ class InputStreamProxy:
     
     def connect(self, output):
         if isinstance(output, dict):
+            print('InputStreamProxy.connect', output)
             self.node.connect_input(self.name, output)
         else:
+            print('InputStreamProxy.connect', output.params)
             self.node.connect_input(self.name, output.params)
         
         
--- a/pyacq/core/client.py
+++ b/pyacq/core/client.py
@@ -81,6 +81,7 @@ class NodeProxy(object):
         self.inputs = { name:InputStreamProxy(self, name) for name in self.nodegroup.get_node_attr(self.name, '_input_specs').keys()}
         self.outputs = { name:OutputStreamProxy(self, name) for name in self.nodegroup.get_node_attr(self.name, '_output_specs').keys()}
         
+        
     def __getattr__(self, name):
         return lambda *args, **kwargs: getattr(self.nodegroup, 'control_node')(self.name, name, *args, **kwargs)
 
@@ -113,5 +114,14 @@ class InputStreamProxy:
         self.name = name
     
     def connect(self, output):
-        self.node.connect_input(self.name, output.params)
+        if isinstance(output, dict):
+            self.node.connect_input(self.name, output)
+        else:
+            self.node.connect_input(self.name, output.params)
+        
+        
+
+    @property
+    def params(self):
+        return self.node.get_output(self.name)
 
--- a/pyacq/core/stream.py
+++ b/pyacq/core/stream.py
@@ -246,6 +246,7 @@ class PlainDataSender:
         return index, data
     
     def send(self, index, data):
+        #TODO deal when data is nparray and self.copy is False and a.flags['C_CONTIGUOUS']=False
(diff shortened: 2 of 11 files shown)


pyacq/core/client.py
```diff
@@ -78,13 +78,21 @@ class NodeProxy(object):
...
-        self.outputs = { name:OutputStreamProxy(self, name) for name in self.nodegroup.get_node_attr(self.name, '_output_specs').keys()}
+        #~ self.inputs = { name:InputStreamProxy(self, name) for name in self.nodegroup.get_node_attr(self.name, '_input_specs').keys()}
+        #~ self.outputs = { name:OutputStreamProxy(self, name) for name in self.nodegroup.get_node_attr(self.name, '_output_specs').keys()}
         
     def __getattr__(self, name):
         return lambda *args, **kwargs: getattr(self.nodegroup, 'control_node')(self.name, name, *args, **kwargs)
 
     @property
+    def inputs(self):
+        return { name:InputStreamProxy(self, name) for name in self.get_input_names() }
```
campagnola: Do we really want new `InputStreamProxy` instances created every time this is called?

pyacq/core/stream.py
```diff
@@ -80,9 +82,14 @@
 class OutputStream:
     """A OutputStream is a helper class to send data.
     """
-    def __init__(self, spec = {}):
+    def __init__(self, spec = {}, node = None, name = None):
         self.configured = False
         self.spec = spec # this is a priori stream params, and must be change when Node.configure
+        if node:
```
campagnola: Is a Node true or false? Is None true or false? I can never remember.
Better to be explicit:  `if node is None`

pyacq/core/client.py
samuelgarcia: At the beginnings it was not the case (done at instanciation) but some Nodes have number of INput/ouput which is dynamic (ex Splitter) in that case we need to call RPC method to get the real input/output list. It is not the standart usecase but it happens.

Do you have better ideas ?

pyacq/core/stream.py
samuelgarcia: OK.

pyacq/core/client.py
campagnola: I see.  They could be cached, but it's not terribly important (I was just curious).

pyacq/core/tools.py
```diff
@@ -81,28 +114,111 @@ def _initialize(self):
...
+            if self.input.params['timeaxis']==0:
+                assert output.params['shape'][1] == 1, 'StreamSplitter: wrong shape'
+            else:
+                assert output.params['shape'][0] == 1, 'StreamSplitter: wrong shape'
+
+    def after_input_connect(self, inputname):
+        if self.input.params['timeaxis']==0:
+            self.channelaxis = 1
+            self.nb_channel = self.input.params['shape'][1]
+            shape = (-1, 1)
```
campagnola: We seem to do a lot of work to handle the axis order.. What do you think about just requiring all streams to have time on axis 0?

pyacq/viewers/qtimefreq.py
```diff
@@ -0,0 +1,739 @@
...
+        
+        #~ self.last_wt_map = wt
+        self.out_stream().send(head, wt)
+        #~ t3 = time.time()
+        
+        #print('compute', self.channel,  t2-t1, t3-t2, t3-t1, QtCore.QThread.currentThreadId())
+
+
+
+class TimeFreqWorker(Node, QtCore.QObject):
```
campagnola: What do you think about adding a third node category like `device`, `viewer`, `filter` ? I seem to recall we discussed this but I don't remember what we decided. In any case, this node would be useful independent of `QTimeFreq`.

pyacq/viewers/qtimefreq.py
```diff
@@ -0,0 +1,739 @@
...
+    Class to visulaize time-frequency with Morlet continuous wavelet.
+    
+    It allow better visualisation the standart FFT spectrogram because it is not dependant
+    of the FFT size along frequency.
+    See https://en.wikipedia.org/wiki/Morlet_wavelet
+    
+    The implementation is done with one TimeFreqWorker per channel.
+    The computation is quite heavy:
+        * downsampling the signal chunk (with a filtfilt first)
+        * fft of signal
```
campagnola: It might not be obvious to readers that fft-multiply-ifft results in a wavelet convolution..

pyacq/viewers/qtimefreq.py
```diff
@@ -0,0 +1,739 @@
...
+        * multiplication line by line (freq by freq) with the FFT wavelet.
+        * ifft the maps
+        
+    There is 2 modes:
+        * each TimeFreqWorker live in the same QApp as QTimeFreq (nodegroup_friends=None)
+        * each TimeFreqWorker is spawn in some other NodeGroup to dustrubute the  (nodegroup_friends=[some_list_of_nodegroup])
```
campagnola: This is wicked cool :)

pyacq/core/tools.py
samuelgarcia: Yes it is not elegent maybe we could do some method to help at stream level.
Axis 0 for all :No. This is really important to have choice between channel C-continuity or time C-continuity to avoid mem copy when doing FFT.

pyacq/viewers/qtimefreq.py
samuelgarcia: At the moment there is not special category of Node except the dirname they are.

Yes this TimeFreqWorker could be somewhere else. We will see when re implementing with vispy.

pyacq/viewers/qtimefreq.py
samuelgarcia: tiss tiss tisss. You miss some course in signal processing. :)
I will add some more explicit doc about the method.

pyacq/core/tools.py
campagnola: But the order of axes is independent of the actual memory layout of the data..  Could we make it that streams can send  data with any axis-order, but the result is always transposed to have time=0? That way you have the freedom to choose whichever memory layout is most efficient, but the streams make it appear as if time is always axis 0.  I'll add this to my list of "maybe-do" things and let you decide if it is worthwhile.

pyacq/viewers/qtimefreq.py
samuelgarcia: Thanks but at the moment. We have limitation due to RPC. this second mode imply that QTimeFreq is in local QApp and connot be remoted because nodegroup_friends cannot be Jsonified.
But TimeFreqWorker can be on another machine and this is amazing. 64 channels at 20kHz 2FPS and no lag!!!

pyacq/core/tools.py
samuelgarcia: I was thinking that in C-ordered continuous ndarray : the last axis was continous. Am I wrong ?

I prefer to discuss it with you before implementing something.
I think you better start by continuing RPC stuff. No ?

pyacq/viewers/qtimefreq.py
campagnola: I have "RPC object proxy" in my maybe-do list as well; I'll keep this in mind as a use case.

samuelgarcia: If you start about RPC. I also would like to be able to modify remotly pg.parametertree.Parameter.
Usecase: distant oscilloscope and local OscilloscopeControler that modify Oscilloscope.params and Oscilloscope.by_channel_params

pyacq/core/tools.py
campagnola: > I was thinking that in C-ordered continuous ndarray : the last axis was continous. Am I wrong ?

You are correct, but ndarray does not have to be C-contiguous. Calling `array.transpose(...)` does not actually copy the memory, it just returns a new ndarray pointing to the same memory but with strides swapped.

pyacq/viewers/qtimefreq.py
campagnola: Been pondering this. I am wondering whether it is more work to bring object proxies to pyacq's RPC or bring zmq to pg.multiprocess...

pyacq/core/tools.py
samuelgarcia: In our cased, it is a SharredArray created by the OutputStream that do it. So we are in that case.

pyacq/viewers/qtimefreq.py
samuelgarcia: I have think a bit about this and not have fix solution. Make the good choice.
Porting zmq in pg could be really cool but it more work.

PR #25, opened 2015-10-07
Testing for travis errors

campagnola:
(do not merge)

### DIFF
--- a/pyacq/core/__init__.py
+++ b/pyacq/core/__init__.py
@@ -7,3 +7,4 @@ from .nodelist import register_node_type
 from .manager import Manager, create_manager
 from .stream import OutputStream, InputStream
 from .sharedarray import SharedArray
+from .tools import ThreadPollInput
--- a/pyacq/core/node.py
+++ b/pyacq/core/node.py
@@ -6,7 +6,7 @@ from .stream import OutputStream, InputStream
 from logging import info
 
 
-class Node(QtCore.QObject):
+class Node(object):
     """
     This class:
         * is a basic element to process data stream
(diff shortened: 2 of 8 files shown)


samuelgarcia:
When need to handle QSignal in a Node is it possible to have QSignal with object.
I propose:
BaseNode(object)
Node(BaseNode, QObjetc)
NodeWidget(BaseNode, QWidget)

campagnola:
Are there any nodes that emit signals? 

In the past I have handled it like this:

```
class MyClass(object):
    # Only used for emitting signals:
    class SignalProxyObject(QtCore.QObject):
        my_signal = QtCore.Signal()
    
    def __init__(self):
        self.__sig_proxy = MyClass.SignalProxyObject()
        self.my_signal = self.__sigProxy.my_signal
```

samuelgarcia:
Yes there are some nodes in all PR I made. I will have a look.
I finally propose:
Node(object)
QNode(Node, QObject)
QWidgetNode(Node, QWidget)

So each can inherits what it needs.

campagnola:
If the only purpose of QNode is to introduce QObject as a superclass, then why not just do that for the subclasses that require it? 

Instead of this:

```
Node(object)
QNode(Node, QObject)
MyQNode(QNode)
```

Why not this:

```
Node(object)
MyQNode(Node, QObject)
```

samuelgarcia:
I would like at least QWidgetNode because close is more tricky (2 meaning :  QWidget.close and Node.close).

campagnola:
Agreed.

samuelgarcia:
OK so I mergel this patch in my samuel.garcia:device-pyaudio
And pull it in all my branches to see what append.
OK ?

campagnola:
+1

PR #29, opened 2015-10-10
Add pyaudio_tfr example

campagnola:
Just a simple example for testing tfr.

### DIFF
--- a/example/pyaudio_tfr.py
+++ b/example/pyaudio_tfr.py
@@ -23,33 +23,35 @@ ng = man.create_nodegroup()
 # Create PyAudio device node in remote process
 dev = ng.create_node('PyAudio')
 
-# Print a list of available input devices (but ultimately we will just use the 
-# default device).
-default_input = dev.default_input_device()
-print("\nAvaliable devices:")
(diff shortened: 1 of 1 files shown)


example/pyaudio_tfr.py
```diff
@@ -0,0 +1,64 @@
...
+              format='int16', chunksize=1024)
+dev.output.configure(protocol='tcp', interface='127.0.0.1', transfertmode='plaindata')
+dev.initialize()
+
+
+
+# Create a viewer in the local application, using the remote process for
+# frequency analysis
+viewer = QTimeFreq()
+viewer.configure(with_user_dialog=True, nodegroup_friends=[ng])
```
samuelgarcia: nodegroup_friends = [ man.create_nodegroup() for i in range(4) ]
is more efficient in multi core/cpu machine.
this example could be more specific about that.

The example could also explain that for device Node a good pratice is let it alone in its NodeGroup (process).

campagnola: In this case there is only one channel, so I don't think we will see any performance gain? I could at least add some comments..

PR #30, opened 2015-10-10
Englishify

campagnola:
Filled out docstrings, corrected for English, and applied some PEP8 style rules. 
The commits are independent, so I can revert parts of this PR if you disagree with any changes..


samuelgarcia:
Thank you so much for is enormous clean.
Shame on me for all dirty docs.

The only strange thing to me is the replacement of "sampling_rate" to "sample_rate".
I never known the good usage. Is there a diff between US and UK ?
But I don't care.

campagnola:
I have always used "sample rate" exclusively. Upon closer examination it appears that the two are used with about equal frequency.

PR #33, opened 2015-10-16
Split rpc into subpackage

campagnola:
Just dividing up the rpc code here because it's about to get more complicated..


samuelgarcia:
We still have the error in travis but this come from timefreq, so I merge anyway.

PR #34, opened 2015-10-19
0.1.x

Lx37:
add windows case to emotiv device


samuelgarcia:
need to be rewritten with pyacq0.2 version.

samuelgarcia:
need to be rewritten with pyacq0.2 version.

PR #35, opened 2015-10-23
Object proxy

campagnola:
Finally have a working object-proxying RPC system. This has not yet been fully integrated, but now seems like a good time to stop, look at what we have, and decide how to proceed. The consequences of this design are much deeper than I had previously realized. Almost all of the work previously done by Host, NodeGroup, and Manager are now absorbed into the RPC system, leaving these classes rather bare (and possibly unnecessary). See `NodeGroup` and `Host` for example; I haven't worked on `Manager` yet.

I've added a new example `pyaudio_oscope2.py` that demonstrates acquiring audio data in a separate process and displaying it locally. It is almost identical to the original `pyaudio_oscope.py`. 

To move on, we will need to decide whether to keep the old management classes and what their new purposes will be.

*Update:*

* Manager, Host, and NodeGroup have very close to the same functionality as before (but with much less code).
* All log messages are collected across hosts, processes, and threads and combined at the manager.
* Additionally, stdout/stderr lines and unhandled exceptions are collected from all spawned processes and converted to log messages.
* Log messages are sorted in time before they are printed such that the log records actually have a readable sequence (but this will require precise time synchronization between hosts; perhaps we can devise some network - or hardware-based strategy for this).
* RPC requests re fully reentrant. For example, the manager may call a remote method on a Node, which directly calls back to the manager again.
* RPC server can run in its own thread or in a Qt GUI thread


samuelgarcia:
Waou. Awsome. this is a huge change!
I will need time to read and understand everything.

Before I read everything, could you also port test_node.py and example/local_or_remote.py.
So I am sure to not misanderstand.

samuelgarcia:
Maybe you have push your latest ProcessSpawner I have this error :
AttributeError: 'ProcessSpawner' object has no attribute '_import'

samuelgarcia:
OK. I read the main part of this.
Globaly I like this, this is pyqtgraph approach recoded with zmq.
I still have some doubt maybe we coul discuss this a phone call.

Here my very first feelings:
  * Are yousSure to leave this in pyacq and not in pyqtgraph ?

  * Are you sure that this new RPC is not too much python centric ?
    How easy will it be to write some client and viewer nodes in javasciprt for instance ?
    The old approach:
        * pros: 
            * easly portable to other language
        * contr:
             * very verbose. need to write both server/client(proxy) side
             * many methods at NodeGroup level.
     The new approach: the opposite.

  * Could you make some drafty benchmark:
        * how many per second : ping/pong (on local host)
        * ow many per second  : simple metod that return None (on local host) synchronously

    * In NodeGroup : nodes =set() : how do we ensure unique name for nodes ? I guess it is only at manager level

    * What about JSON ? Can we have choice at ProcessSpawner level ?

    * QtPollThread.rpc_socket = server._socket
Is this safe ? because the thread access an attribute of its parent ? If yes I think that I put weakref everywhere for nothing. tell me.

   * Why this : ObjectProxy.__reduce__ ? I guess you forgiot to remove. Are you palnning to use pickle ?

    * RPCClient.process seems to be buggy (and unused??) because it call self.process_msg(name, msg)

campagnola:
> Are yousSure to leave this in pyacq and not in pyqtgraph ?

I was thinking that it might make a nice standalone library. Perhaps we can think about publishing it on its own sometime.

> Are you sure that this new RPC is not too much python centric ?
> How easy will it be to write some client and viewer nodes in javasciprt for instance ?

I think, from a language-agnostic standpoint, this version is only a little more complicated than the last. Before we had to pass messages like 

    {'action': 'call', 'req_id': 1, 'method': 'start', 'args': [...], 'kwargs': {...}}

Now the only differences are that there are a few more actions ('get_obj' being the most important) and you have to keep track of proxies like `{'rpc_id': 'tcp://127.0.0.1:5436', 'obj_id': 42}`. From this standpoint, the proxy approach is more complex/powerful, but not any less language-agnostic. 

It does look bit more Pythonic when I say

    dev_proc = rpc.ProcessSpawner()
    dev = dev_proc.client._import('pyacq.devices.audio_pyaudio').PyAudio()

..but this could be very easily wrapped in a way that does not require `_import`. Before `NodeGroup` and `Manager` shared this task of creating nodes based on their class name; we can bring this back with just a few lines of code.

> Could you make some drafty benchmark:

Will work on these..

> In NodeGroup : nodes =set() : how do we ensure unique name for nodes ? I guess that unicity is only at manager level

Previously we used these names to reference objects across machines/processes, and we needed the manager to keep track of the names to ensure they were unique. Now we reference all objects using ObjectProxy, which is always guaranteed to be unique. This means that it is no longer necessary to name these objects (although it might be helpful for logging).

> What about JSON ? Can we have choice at ProcessSpawner level ?

We can, I just haven't implemented it yet. I started with msgpack because it is more efficient, especially for transferring array data. My temptation would be to have RPCClient decide what message format it wants to use, and then RPCServer can adapt accordingly. That way we can use the most efficient format internally, while still allowing MATLAB/others to connect by JSON if they want.

> QtPollThread.rpc_socket = server._socket
> Is this safe ? because the thread access an attribute of its parent ? If yes I think that I put weakref
> everywhere for nothing. tell me.

It is safe because just below that I have `server._socket = context.socket(zmq.PAIR)`. The original socket becomes unavailable to the server and belongs to the poller thread instead. 

> Why this : ObjectProxy.reduce ? I guess you forgiot to remove. Are you palnning to use pickle ?

ObjectProxy was originally just copied from pyqtgraph. This is just left-over. 

> RPCClient.process seems to be buggy (and unused??) because it call self.process_msg(name, msg)

Left-over again. I'll remove these.

campagnola:
Added a benchmarking script. Results on my laptop:

```
=========== inproc to thread ============
Ping rate: 4089/sec
Async ping rate: 4616/sec
=========== tcp to thread ============
Ping rate: 2972/sec
Async ping rate: 3987/sec
=========== TCP to spawned process ============
Ping rate: 2982/sec
Async ping rate: 16614/sec
=========== TCP to spawned Qt process ============
Ping rate: 2260/sec
Async ping rate: 6219/sec
```

samuelgarcia:
This is cool. It is a bit slower than the previous RPC but not so much.

campagnola:
Updated description at the top of the PR.

pyacq/core/host.py
```diff
@@ -1,62 +1,66 @@
...
         
-        Return the RPC name and address of the new nodegroup.
-        """
-        assert name not in self.nodegroup_process, 'This node group already exists'
-        # print(self._name, 'start_nodegroup', name)
-        ps = ProcessSpawner(NodeGroup, name, addr)
-        self.nodegroup_process[name] = ps
-        return ps.name, ps.addr
-    
-    def close_nodegroup(self, name, force=False):
```
samuelgarcia: What is the new way of closing a NodeGroup ?

samuelgarcia:
class Node have not been changed by the PR. Is it a missing ?
I think many methods there can be removed:
get_input_names, get_output_names, connect_input, configure_output, get_output, get_input

pyacq/core/manager.py
```diff
@@ -22,17 +23,40 @@ def create_manager(mode='rpc', auto_close_at_exit=True):
...
+    while len(root_logger.handlers) > 0:
+        root_logger.removeHandler(root_logger.handlers[0])
+    root_logger.addHandler(log_handler)
+    
+    # Send local uncaught exceptions through logger for nice formatting
+    rpc_log.log_exceptions()
+        
+    # start a global log server
+    if rpc_log.get_logger_address() is None:
+        rpc_log.start_log_server(logger)
```
samuelgarcia: This is cool!

pyacq/core/manager.py
```diff
@@ -47,188 +71,127 @@ class Manager(RPCServer):
...
+                raise TimeoutError("No response from host at %s" % addr)
+                
+            try:
+                self.hosts[addr] = cli['host']
+            except KeyError:
+                raise ValueError("Contacted %s, but found no Host there." % addr)
+        return self.hosts[addr]
+    
+    def disconnect_host(self, host):
+        host.close_nodegroups(self)
```
samuelgarcia: Is it : Host.close_all_nodegroups instead ?

pyacq/core/manager.py
```diff
@@ -47,188 +71,127 @@ class Manager(RPCServer):
...
-        RPCServer.__init__(self, name, addr)
-        
-        self.hosts = {}  # name:HostProxy
-        self.nodegroups = {}  # name:NodegroupProxy
-        self.nodes = {}  # name:NodeProxy
-        
-        # auto-generated host on the local machine
-        self._default_host = None
+        # Host used for starting nodegroups on the local machine
+        self.default_host = Host('default_host')
```
samuelgarcia: So default Hiost and manager live the same process, this is much simpler. Thanks.

pyacq/core/tests/test_host.py
```diff
@@ -2,44 +2,28 @@
...
-    
-    time.sleep(1.)
-    
-    process_host0.stop()
-
+    p1.stop()
+    p2.stop()
 
 
 if __name__ == '__main__':
```
samuelgarcia: test_host2() dp not exists anymore.

pyacq/core/rpc/tests/rpc_benchmark.py
```diff
@@ -0,0 +1,68 @@
...
+test_poingrate(proc.client)
+test_async_poingrate(proc.client)
+
+
+# process
+print("=========== TCP to spawned Qt process ============")
+proc = rpc.ProcessSpawner(qt=True)
+test_poingrate(proc.client)
+test_async_poingrate(proc.client)
+
```
samuelgarcia: On my machine I get this:

=========== inproc to thread ============
Ping rate: 2644/sec
Async ping rate: 2786/sec
=========== tcp to thread ============
Ping rate: 1672/sec
Async ping rate: 2697/sec
=========== TCP to spawned process ============
Ping rate: 1652/sec
Async ping rate: 4687/sec
=========== TCP to spawned Qt process ============
Ping rate: 1268/sec
Async ping rate: 4396/sec
WARNING:pyacq.core.rpc.server:RPCServer exiting without close()!
Error in atexit._run_exitfuncs:
  File "/home/samuel/Documents/projet/pyacq-0.2/pyacq/core/rpc/client.py", line 306, in _read_and_process_all
    self._read_and_process_one(timeout=0)
  File "/home/samuel/Documents/projet/pyacq-0.2/pyacq/core/rpc/client.py", line 294, in _read_and_process_one
    self._socket.setsockopt(zmq.RCVTIMEO, timeout)
  File "zmq/backend/cython/socket.pyx", line 308, in zmq.backend.cython.socket.Socket.set (zmq/backend/cython/socket.c:3163)
  File "zmq/backend/cython/socket.pyx", line 96, in zmq.backend.cython.socket._check_closed (zmq/backend/cython/socket.c:1324)
zmq.error.ZMQError: Socket operation on non-socket
WARNING:pyacq.core.rpc.server:RPCServer exiting without close()!
Error in atexit._run_exitfuncs:
  File "/home/samuel/Documents/projet/pyacq-0.2/pyacq/core/rpc/client.py", line 306, in _read_and_process_all
    self._read_and_process_one(timeout=0)
  File "/home/samuel/Documents/projet/pyacq-0.2/pyacq/core/rpc/client.py", line 294, in _read_and_process_one
    self._socket.setsockopt(zmq.RCVTIMEO, timeout)
  File "zmq/backend/cython/socket.pyx", line 308, in zmq.backend.cython.socket.Socket.set (zmq/backend/cython/socket.c:3163)
  File "zmq/backend/cython/socket.pyx", line 96, in zmq.backend.cython.socket._check_closed (zmq/backend/cython/socket.c:1324)
zmq.error.ZMQError: Socket operation on non-socket

pyacq/core/rpc/tests/test_rpc.py
```diff
@@ -0,0 +1,347 @@
+import threading, atexit, time, logging
```
samuelgarcia: I get this on my machine:
[22m[32m[samuel-HP-Z800-Workstation:main_process:thread-7f730ec25700][0m [1m[33mFailed to call object <bound method TestClass.add of <__main__.test_rpc.<locals>.TestClass object at 0x7f731a953e10>>: 2, ('x',)[0m
[22m[32m[samuel-HP-Z800-Workstation:main_process:thread-7f730ec25700][0m [1m[33m    => returning exception for 18: (<class 'TypeError'>, TypeError("unsupported operand type(s) for +: 'int' and 'str'",), <traceback object at 0x7f73197fc1c8>)[0m
  File "test_rpc.py", line 348, in <module>
    test_disconnect() 
  File "test_rpc.py", line 312, in test_disconnect
    server_proc = ProcessSpawner()
  File "/home/samuel/Documents/projet/pyacq-0.2/pyacq/core/rpc/processspawner.py", line 164, in __init__
    status = bootstrap_sock.recv_json()
  File "/usr/lib/python3/dist-packages/zmq/sugar/socket.py", line 412, in recv_json
    msg = self.recv(flags)
  File "zmq/backend/cython/socket.pyx", line 631, in zmq.backend.cython.socket.Socket.recv (zmq/backend/cython/socket.c:5799)
  File "zmq/backend/cython/socket.pyx", line 665, in zmq.backend.cython.socket.Socket.recv (zmq/backend/cython/socket.c:5599)
  File "zmq/backend/cython/socket.pyx", line 139, in zmq.backend.cython.socket._recv_copy (zmq/backend/cython/socket.c:1752)
  File "zmq/backend/cython/checkrc.pxd", line 21, in zmq.backend.cython.checkrc._check_rc (zmq/backend/cython/socket.c:6275)
zmq.error.ZMQError: Appel système interrompu
Error in atexit._run_exitfuncs:
  File "/home/samuel/Documents/projet/pyacq-0.2/pyacq/core/rpc/processspawner.py", line 188, in stop
    closed = self.client.close_server()
AttributeError: 'ProcessSpawner' object has no attribute 'client'

pyacq/core/rpc/tests/test_serializer.py
```diff
@@ -0,0 +1,28 @@
...
+                g=datetime.date(2015, 1, 1),
+                h=(1,2),
+                #i=[1,2],  # msgpack only returns tuples, not lists (because we have use_list=False).
+                           # see: https://github.com/msgpack/msgpack-python/issues/98
+                )
+    
+    #serializers = [JsonSerializer()]
+    #if HAVE_MSGPACK:
+        #serializers.append(MsgpackSerializer())
+    serializers = [MsgpackSerializer()]
```
samuelgarcia: Is it a lot of work having also JSON ?

samuelgarcia:
This is really good!!
I think, TODO before merging : 
1. fix some bugs (see some coments)
2. JSON again ?
3. clean class Node

pyacq/core/rpc/server.py
```diff
@@ -18,93 +28,465 @@ class RPCServer(object):
...
+    a timer to poll the RPC socket. Responses are sent back to the poller
+    thread by a secondary socket.
+    """
+    new_request = QtCore.Signal(object, object)  # client, msg
+    
+    def __init__(self, server):
+        # Note: QThread behaves like threading.Thread(daemon=True); a running
+        # QThread will not prevent the process from exiting.
+        QtCore.QThread.__init__(self)
+        self.server = server
```
samuelgarcia: Is this safe, in the point of view that a Qt python object must not have pointer to its parents ?

pyacq/core/rpc/client.py
```diff
@@ -1,89 +1,428 @@
...
+            req_id = None
+        else:
+            req_id = self.next_request_id
+            self.next_request_id += 1
+        cmd['req_id'] = req_id
+        logger.info("RPC request '%s' to %s [req_id=%s]", cmd['action'], 
+                    self.address.decode(), req_id)
+        logger.debug("    => %s", cmd)
+        
+        # double-serialize opts to ensure that cmd can be read even if opts
```
samuelgarcia: This I don't anderstand....

pyacq/core/host.py
campagnola: `nodegroup.close()`   :D

pyacq/core/manager.py
campagnola: It's really nice for debugging, especially because after it collects log records from all other processes, it sorts them by time before printing. Only issue is that different machines will have clock differences. I have a function to measure the clock diff, but haven't done anything with it yet..

pyacq/core/manager.py
campagnola: No--a Host can be used simultaneously by more than one Manager, so we ask it to close only the nodegroups that belong to this Manager.

pyacq/core/tests/test_host.py
campagnola: I took it out because everything in that test was covered in the nodegroup and manager tests.

pyacq/core/rpc/tests/rpc_benchmark.py
campagnola: Fixed, and also improved performance. The async-tcp-process benchmark is back up to 16kHz on my machine.

pyacq/core/rpc/tests/test_rpc.py
campagnola: On OSX? I can run some tests there..

pyacq/core/rpc/server.py
campagnola: It is safe--the QThread does not have a parent.

pyacq/core/rpc/client.py
campagnola: Let's say I put an object into `opts` that causes an error during unserialization. This would cause `req_id` to become unreadable, which would prevent sending back an error message. So the double serialization guarantees that we at least get `req_id` before moving on to the more dangerous `opts`. However, we can probably accomplish the same using multipart messages (and it might be faster as well)...  I will investigate.

campagnola:
* Added JSON client support
* Cleaned up Node class

Remaining issues:
* Fix rpc test bug
* Try to avoid double serialization

pyacq/core/manager.py
samuelgarcia: Waouh. Much sofisticated than I thougth.

pyacq/core/rpc/tests/rpc_benchmark.py
samuelgarcia: yeeh!

pyacq/core/rpc/tests/test_rpc.py
samuelgarcia: Me never!
On ubuntu.

samuelgarcia:
I don't anderstand this double serilization.
Could you explain ?

campagnola:
see https://github.com/pyacq/pyacq/pull/35#discussion_r45834212

pyacq/core/rpc/client.py
samuelgarcia: OK I get it.
multipart sound good.

samuelgarcia:
Luke : You can merge when want (since travis get back to green!).
Congratulation for that PR.

campagnola:
Ok, I think this is ready! The only travis error is the size check, so we should be safe to merge. Anyone like to press the button?

PR #36, opened 2015-10-29
Viewer triggeroscilloscope

samuelgarcia:
This PR include:
  * AnalogTrigger (pure python) maybe optimize this one day.
  * TriggerAccumulator
  * TriggeredOscilloscope

Note the difficult was the debounce engine in the trigger.


pyacq/core/tools.py
```diff
@@ -52,13 +52,22 @@ def process_data(self, pos, data):
...
     def stop(self):
-        with self.lock:
+        with self.running_lock:
             self.running = False
     
     def pos(self):
         return self._pos
 
-
+class ThreadPollOuput(ThreadPollInput):
```
campagnola: "Output"

pyacq/dsp/tests/test_triggeraccumulator.py
```diff
@@ -0,0 +1,91 @@
...
+from pyqtgraph.Qt import QtCore, QtGui
+
+
+nb_channel = 2
+sample_rate =1000.
+chunksize = 100
+
+length = int(sample_rate*20)
+t = np.arange(length)/sample_rate
+buffer = np.random.rand(nb_channel, length)*.3
```
campagnola: set the random seed or else we might get random test failures.

pyacq/core/tools.py
samuelgarcia: Yep maybe the name is not good.
The idea is same as ThreadPollinput (that pool a client socket InputStream) but you give it an OuputStream (server socket) so it create internally it is own InputStream.
Maybe you could help with a good english to describe this.
the code uis more clear than the name.

pyacq/dsp/tests/test_triggeraccumulator.py
samuelgarcia: In that case, it does not impact the test. Is is just to have a fake noise chen debuging and watching the oscope.
But agree for Node that really depend random.

campagnola: Ah, I see. I thought there would be an assert somewhere..

samuelgarcia: I am stupid. This random do impact the test. because the thresh hold is 1.
I ptach this.

pyacq/core/tools.py
campagnola: I think the name is ok. I just meant that "Output" is misspelled.

How about this for a docstring:

```
Thread that monitors an OutputStream in the background and emits a Qt signal
when data is sent.

Like ThreadPollInput, this class can be used where low-latency response to data
is needed within a Qt main thread (because polling from the main thread with
QTimer either introduces too much latency or consumes too much CPU).
    
The `process_data()` method may be reimplemented to define other behaviors.
"""

```

samuelgarcia:
OK for merging this ?

PR #38, opened 2015-11-13
Core devel

samuelgarcia:
This try to simplify the timeaxis mess.
Internally timaxis has the meaning as before (in C continuous).
Now by default OuputStream.send(autoswapaxes=True) and InputStream.recv(autoswapaxes=True) exepct a numpy array with timeaxis=0, internally a data.swapaxes in done when necessary.
So Nodes only play with data where dim0 is time even if it is not internally continuous.


pyacq/core/stream.py
```diff
@@ -284,10 +309,14 @@ def compress_blosclz4(self, index, data):
...
-    def send(self, index, data):
-        # TODO deal when data is nparray and self.copy is False and a.flags['C_CONTIGUOUS']=False
-        for f in self.funcs:
-            index, data = f(index, data)
+    def send(self, index, data, autoswapaxes = True):
+        if isinstance(data, np.ndarray):
+            if autoswapaxes and self.params['timeaxis']!=0:
+                data = data.swapaxes(0, self.params['timeaxis'])
+            if not data.flags['C_CONTIGUOUS']:
+                data = data.copy()
```
campagnola: Could be expensive.. why is this necessary?

pyacq/core/stream.py
```diff
@@ -309,20 +338,24 @@ def __init__(self, socket, params):
         elif self.params['compression'] in ['blosc-blosclz', 'blosc-lz4']:
             self.func = self._uncompress_blosc
     
-    def recv(self):
+    def recv(self, autoswapaxes = True):
         m0,m1 = self.socket.recv_multipart()
         index = np.fromstring(m0, dtype='int64')[0]
-        return self.func(index, m1)
+        return self.func(index, m1, autoswapaxes = autoswapaxes)
```
campagnola: Can the axis-swapping code be placed here instead of in `_numpy_fromstring` and `_uncompress_blosc` to avoid repetition?

pyacq/core/stream.py
samuelgarcia: With th autoswap axis we need to ensure a continuity.
Maybe this not necessary because I think, that zmqsocket.send(numpy_array) perform a tostring.
And by default  array.tostring(order='C') so it should be OK.
I did not make test. I should.

pyacq/core/stream.py
samuelgarcia: Yes of course.

campagnola: Better yet, it could probably be moved to `InputStream.recv`..

campagnola: I think this would also avoid the need to pass `autoswapaxes` around.

samuelgarcia: Yep.

pyacq/core/stream.py
campagnola: Looks like `sock.send(array)` will accept arrays that are either c-contiguous or f-contiguous. Anything else will generate an exception, so this check is necessary, but you could change the condition to `if not (data.flags['C_CONTIGUOUS'] or data.flags['F_CONTIGUOUS'])`.

pyacq/core/stream.py
```diff
@@ -284,10 +309,14 @@ def compress_blosclz4(self, index, data):
...
-            index, data = f(index, data)
+    def send(self, index, data, autoswapaxes = True):
+        if isinstance(data, np.ndarray):
+            if autoswapaxes and self.params['timeaxis']!=0:
+                data = data.swapaxes(0, self.params['timeaxis'])
+            if not data.flags['C_CONTIGUOUS']:
+                data = data.copy()
+            for f in self.funcs:
+                index, data = f(index, data)
         self.socket.send_multipart([np.int64(index), data], copy=self.copy)
```
campagnola: Above you set `self.copy = self.params['protocol'] == 'inproc'`, but we can get a performance boost by using no-copy with tcp as well.

pyacq/core/stream.py
```diff
@@ -309,20 +338,24 @@ def __init__(self, socket, params):
         elif self.params['compression'] in ['blosc-blosclz', 'blosc-lz4']:
             self.func = self._uncompress_blosc
     
-    def recv(self):
+    def recv(self, autoswapaxes = True):
         m0,m1 = self.socket.recv_multipart()
         index = np.fromstring(m0, dtype='int64')[0]
```
campagnola: You can get a performance boost here by using `np.frombuffer` instead (it will avoid a copy).

campagnola: Strike that, it's just the index..

pyacq/core/stream.py
samuelgarcia: What append if 'F_CONTINOUS' the receivers get a transposed array no ?
We have on the streamspec one degree of freedom  (for the axis) If we had a second possibility to have C or F this make things more complicated.
The receiver perform a np.frombuffer so it needs to known everything (axis and F/C) and this has to fixed in advance. I reject F_CONINUOUS to simplify. WHat do you think ?

pyacq/core/stream.py
samuelgarcia: Does it works ?

pyacq/core/stream.py
campagnola: Right, if you want to reconstruct F_CONTIG on the remote end, then you also have to set the strides. It's more effort, but might avoid a copy in some cases..  fine by me to leave it C-CONTIG only for now.

pyacq/core/stream.py
campagnola: In my test it does help. Give it a try and see if your stream tests go any faster..

pyacq/core/stream.py
samuelgarcia: F_CONTINOUS axis=0 is the same as C_CONTINUOUS axis=1 so I think we have the necessary for tunning performences.

pyacq/core/stream.py
samuelgarcia: I will had this too.

campagnola:
I am so confused  :D
I thought this would simplify the viewer nodes, but I still see them checking for `timeaxis`. Now I can't even remember what this PR is supposed to do. I thought the idea was that we would always enforce timeaxis=0 so that nodes never need to worry about the axis order?

samuelgarcia:
I think this simplify a lot Nodes that perform a filter. They don't care wich timeaxis. They always do it on 0. Numpy is working under the hood for us.
The complexity of the viewer do not come from the timeaxis but from the fact they need a transfermode = 'sharrearray' BUT they also accept transermode = 'plaindata' so they are creating a proxy input and this proxy input do care of the timeaxis (I wante 1).

I want to create a class for that and Oscilloscope, and other will ingerits from that mecanisms.

Don't be confused this PR was really usefull otherwise the next one wonuld not be possible (filters... PR#39)

campagnola:
Should I take care of the axis-swap code or are you already working on that?

samuelgarcia:
Hi Luke. I take care of finishing this.
I was very busy last 2 weeks. I hope to be back next week on thsi topic.

pyacq/core/stream.py
samuelgarcia: Performences are stranges:
on my linux box.
copy=True have better performences for (tcp, inproc, ipc) for SMALL buffer for large buffer
copy=False have better performences for large buffer.

Could you check benchmark_stream() in test_stream.py ?
and tell me yoru opinion ?

samuelgarcia:
Have a last eye and tell me.

campagnola:
This looks good. I did a lot of stream performance testing and will post the results in a new PR.. there are still a few things we can do to improve throughput.

PR #45, opened 2015-12-18
Stream performance

campagnola:
This doesn't yet change any behavior of pyacq, but implements some benchmarks that should start us on the path to maximizing stream performance.
- Added `data_transfer_sandbox`, which is a zmq-only demonstration of (I think) the best performance we can expect to see in streams. It minimizes copies and preserves array memory layout (turns out these are mutually dependent goals). This should obviate the need for `autoswapaxes` (if I understand it correctly).
- Moved stream benchmarking to its own file and reworked it a bit such that performance is measured exactly as in the sandbox. Using tcp/plaindata stream gets me ~700MB/s, whereas I can get ~1.3GB/s in the equivalent sandbox test.


campagnola:
@samuelgarcia, I am thinking about giving Streams a little overhaul. Here are the things on my mind:

* Implement data transfer for streams as described above
* Get rid of timeaxis, autoswapaxes, and nb_channel options. 
* Allow inputstream to write data into a circular buffer without re-sending through sharedmem stream

This should simplify the nodes a lot. Can you think of any reasons these should not be done?

samuelgarcia:
Hi luke, I also think about some refactoring about streams and Node instance/init/configure/start/stop/close scenario.

There is a need of simplifying but timeaxis is a need (in my mind at least). autoswapeaxis is a consequence. nb_channel was to simplify the things because it is redundant with shape[X] were X is not the stimeaxis. The ordering of bytes in memory is crucial. We need to cover all possibilities (channel then sample or reversed).

So:
 * +1
 * what do you have in mind.
 * I don't have it exactly understand. Would it be a 'localmem' than would behave exactly like 'sharedmem' but without memmap and without sending the position in the socket ?


I had in a first time somethinng else in mind split Ouput(Intput)Stream in two classes:
  1- one used for connecting Nodes
  2- one (several) for streaming data with multiple scenario (sharedmeme, paliondata in soscket, 'localmem', sharedmeme in GPU, ...)

Now all of theses all intermixed and this is bad. This over complicated Nodes conenction (and the order of connection).

campagnola:
> There is a need of simplifying but timeaxis is a need (in my mind at least). autoswapeaxis is a consequence. nb_channel was to simplify the things because it is redundant with shape[X] were X is not the stimeaxis. The ordering of bytes in memory is crucial. We need to cover all possibilities (channel then sample or reversed).

Why is timeaxis needed? With the changes I proposed above, we can force timeaxis=0 and still preserve the memory order (and plus we get a performance boost for doing this).  

> Would it be a 'localmem' than would behave exactly like 'sharedmem' but without memmap and without sending the position in the socket ?

Yes, something like that. Possibly it could be integrated directly into InputStream so that any stream can have an internal ring buffer and use `get_array_slice`, regardless of the transfer mode (because right now, that effort is placed on the nodes instead).

> I had in a first time somethinng else in mind split Ouput(Intput)Stream in two classes:
> 1- one used for connecting Nodes
> 2- one (several) for streaming data with multiple scenario (sharedmeme, paliondata in soscket, 'localmem', sharedmeme in GPU, ...)

Isn't that exactly how it is implemented now (InputStream/OutputStream are separated from the various Receiver classes) ?

samuelgarcia:
It appear that your doing exactly what I was about to do : make a clear RingBuffer class.
Alway ahead luke!! Do you need help (except reviewing ) ?
I don't anderstand exactly how we will play with timeaxis (0 or 1) but the code is cristal clear. thanks.

Does this code depend on another branch ? there lots of RPC stuff in the commit list.

pyacq/core/stream.py
```diff
@@ -523,11 +576,348 @@ def get_array_slice(self, index, length, autoswapaxes = True):
...
+                                                    "only advanced by %d." % 
+                                                    (dsize, index-self._write_index)) 
+
+        revert_inds = [self._read_index, self._write_index]
+        try:
+            # advance write index. This immediately prevents other processes from
+            # accessing memory that is about to be overwritten.
+            self._set_write_index(index)
+            
+            # decide if any skipped data needs to be filled in
```
samuelgarcia: normally if (index is None) next 5 lines areunessecery no ?
Am I wrong ?
So maybe we could add if index is None: 
This would avoid these lines but need an extra variable _index = index some lines before.
This function will be called heavlly. We must carefull about all checks.

samuelgarcia:
Some comment on the fly while reading:
  * what do you think of a separate file for ringbuffer class ?
  *

pyacq/core/stream.py
```diff
@@ -523,11 +576,348 @@ def get_array_slice(self, index, length, autoswapaxes = True):
...
+                break_index = (self._write_index + 1) - ((self._write_index + 1) % bsize)
+                if (start < break_index) == (stop <= break_index):
+                    start_ind = start % bsize
+                    stop_ind = start_ind + (stop - start)
+                    #print("  single; contig:", (start, stop), (start_ind, stop_ind), bsize, break_index)
+                    data = self.buffer[start_ind:stop_ind]
+                else:
+                    #print("  single; discontig:", (start, stop), (start%bsize, stop%bsize), bsize, break_index)
+                    # need to reconstruct from two pieces
+                    data = np.empty((stop-start,) + self.shape[1:], self.buffer.dtype)
```
samuelgarcia: This case is important.
Maybe we could have 2 behaviors for that case:
   * one doing a memcopy at the discontinuity : the case you implement.
   * one without memcopy that always return data1, data2. data2 is None in the normal case data is not None at the discontinuity. This more annoying for the reader but this avoid copy.

example
```
ringbuffer = RingBuffer(double=False, discontinuitymode='nocopy')
data1, data2 = ringbuffer[ind1:ind2]
process(data1)
if data2 is not None:
  process(data2)


```

pyacq/core/sharedarray.py
```diff
@@ -55,13 +43,62 @@ def __init__(self, shape=(1,), dtype = 'float64', shm_id = None):
...
+        return {'nbytes': self.nbytes, 'shm_id': self.shm_id}
+    
+    def to_numpy(self, offset, dtype, shape, strides):
+        """Return a numpy array pointing to part (or all) of this buffer.
+        """
+        return np.ndarray(buffer=self.mmap, shape=shape,
+                          strides=strides, offset=offset, dtype=dtype)        
+        
+
+
```
samuelgarcia: With the new approch I think we are missing the previous internally timeaxis=1.

Example
```
#Sender side
nb_channel = 16
nb_sample = 1000
sh_arr1 = SharedArray(shape=(nb_channel, nb_sample), dtype='float32', shm_id=None)
np_arr1 = sh_arr1.to_numpy() 
#we are wrong here because timaxis is not 0 so manually:
np_arr1 = np_arr1.swapaxes(0,1)
print(np_arr1.shape) #(1000, 16) OK
#receiver side
spec = pickle.dumps(orig_array.to_dict())
sh_arr2 = SharedArray(**pickle.loads(spec))
np_arr1 = sh_arr2.to_numpy() 
# do not known that we need swapaxes
print(np_arr2.shape) #(16, 1000)  BAD
```

I think we need something more magic No ?
This would lead to add 'swapaxes' key in streamspec.... (which is at the end the generalisation of timeaxis but for ND!!)

```
class SharedArray:
 def __init__(self, shape=(1,), dtype='float64', swapaxes=None, shm_id=None):
        self.shape = shape
        self.dtype = np.dtype(dtype)
        self.swapaxes = swapaxes
        nbytes = np.prod(shape)*self.dtype.itemsize
        self.shmem = SharedMem(nbytes, shm_id)
    
    def to_dict(self):
        return {'shape': self.shape, 'dtype': self.dtype, 'swapaxes':self.swapaxes,  'shm_id': self.shmem.shm_id}
    
    def to_numpy(self):
        np_arr = np.frombuffer(self.shmem.mmap, dtype=self.dtype).reshape(self.shape)
        if self.swapaxes is not None:
            np_arr = np_arr.swapaxes(self.np_arr)
        return np_arr


```

campagnola:
Weird--the RPC changes have already been merged into master. I'm not sure why they are showing up here.

campagnola:
I rebased; that seems to have fixed it.

campagnola:
FYI, these changes are still in mid-overhaul. I am happy to hear your thoughts about the direction it is going, nevertheless..

> what do you think of a separate file for ringbuffer class ?

Definitely, and more. I am planning to add `pyacq/core/stream/` with several new files. (later, to prevent diff confusiuon)

> core.tools.StreamConverter should loose the ability of transposing automatically swap axis by configuring different timaxis on ouputstream_spec

Yes.. I am still somewhat experimenting with this stream code. After that appears to be working, we can start thinking about how it affects the rest of the system. I hope the nodes will become much simpler.

> I guess the goal is to remove timeaxis Given that timeaxis is always 0 now. Internally we have the 2 possibility of (sampleXchannel) and (channelXsample). So in that case we need also to give the strides in SharredArray.to_dict() and SharedMem.to_dict() and in there init() to reconstruct the array as they are. Am I missing something.

You are correct; SharedArray is still broken. 

> This will imply lots of changes in my viewers. Do you need help for that ?

Yes, but not yet. :)

> SharedArraySender/SharedArrayReceiver will need a complete refactor based on RingBuffer do you need help ?

Let me get SharedMemSender/Receiver working first. I'm really not sure how everything should fit together yet.

pyacq/core/sharedarray.py
campagnola: For now, SharedArray is left in as a convenience, but I don't think it will end up being used. RingBuffer makes use of the lower-level SharedMem, which is only a 1-D byte buffer. This will allow more flexibility in how data is ordered when it is written into the buffer. In the end, I hope the user will never need to think about swapaxes--data will pass int a stream and arrive at the other end with the same shape and strides.

samuelgarcia:
Great. I wait. have you seen comments in the code ?

pyacq/core/stream.py
```diff
@@ -523,11 +600,386 @@ def get_array_slice(self, index, length, autoswapaxes = True):
...
+        self.shape = shape
+        
+        shape = (shape[0] * (2 if double else 1),) + shape[1:]
+        # initialize int buffers with 0 and float buffers with nan
+        if fill is None:
+            fill = 0 if np.dtype(dtype).kind in 'ui' else np.nan
+        self._filler = fill
+        
+        bsize = np.product(self.shape) * np.dtype(dtype).itemsize
+        RingBuffer.__init__(self, bsize, double=double, shmem=shmem)
```
samuelgarcia: I known this is a work in progress.
So this is certainly a useless comments.
I have this impression that with 
RingBuffer.__init__(self, bsize, double=double, shmem=shmem)
and this in new chunk
RingBuffer.new_chunk(self, np.ravel(data).view('ubyte'), index*self.dtype.itemsize)
we force the ArrayRingBuffer to have always timeaxis=0 internally (channel are interleaved).

Am I missing something ?

campagnola: No, you're right--ArrayRingBuffer is forced to have timeaxis=0 internally. But then, it does not make any sense to have an array ring buffer that works any other way. If you want to be able to stack array chunks in a ring buffer and be able to extract any segment of the buffer without copy, then you _must_ have timeaxis=0 in the ring buffer.

So my plan is something like:

* Streams send / recv array chunks without altering their internal memory layout--if channels are not interleaved when they are sent, then they will not be interleaved when they are received, regardless of the stream type.
* Sharedmem streams use RingBuffer to store byte-arrays in any axis order
* If you attach an ArrayRingBuffer to an InputStream, then by default it will use its own memory buffer and copy chunks into the buffer with timeaxis=0 as they arrive.
* If a sharedmem stream's buffer already happens to have its memory layout with timeaxis=0, then the ArrayRingBuffer can use the mmap buffer rather than creating its own memory and copying.

samuelgarcia: Outch, I am not sur to follow this.
I really really would like to have the 2 possibilities inetrnally (sample, channel) and (channel, sample). This is one of strong point for me in pyacq 0.2.

> Streams send / recv array chunks without altering their internal memory layout--if channels are not interleaved when they are sent, then they will not be interleaved when they are received, regardless of the stream type.

For *plaindata* this make sens but for sharredarray I would like to have all possibilities:
  1 - sender is (sample, channel) sharredarray is (sample, channel)
  2 - sender is (channel, sample) sharredarray is (sample, channel)
  3 - sender is (sample, channel) sharredarray is (channel, sample)
  4 - sender is (channel, sample) sharredarray is (channel, sample)
Case 2 and 3 imply some tranpose but numpy make it free for us because in any case externally with numpy view everything is (sample, channel). We just have swapaxis in approiate case at array creation.

> Sharedmem streams use RingBuffer to store byte-arrays in any axis order

I would prefer this approach that Sharredmem and SharredArray are agnostic to this ringbuffer mechanism.
But a RingBuffer and RingBuffer array can use sharredmem (optional in __init__), as you start to do.

> If you attach an ArrayRingBuffer to an InputStream, then by default it will use its own memory buffer and copy chunks into the buffer with timeaxis=0 as they arrive.

This is really really good, this was the missing piece in the previous design : a proxy ringbuffer for receiver even if it is not a sharredmem. But I thing we must not force internnally (sample, channel) it must be custumable. But if RingBufferArray have the possibility then we have the feature here.

> If a sharedmem stream's buffer already happens to have its memory layout with timeaxis=0, then the ArrayRingBuffer can use the mmap buffer rather than creating its own memory and copying.

I  think it must be the case event if it is (channel, sample) there are not copy when the stream is sharedmem.
If a receiver needs an internal tranpose then we put a Node in between that just do outbuffer[:] = inbuffer [:].
The internal transpose is done under the wood.

pyacq/core/sharedarray.py
samuelgarcia: I really like the 1D ring buffer.
But I am not so happy with the ND RingBufferArray.
double=True with timeaxis=0 internnaly for 2D array have no sense because the channel is split in 2 part in the memory and then we cannot garanty that all reading are continuous for one channel.

I think this is an advance feature. By default everything should be (channel, sample) internally and externally but for some rare case a swapaxes option is needed. For instance I want to perform lots of FFT on some channel, we really gain to have internnalu continuous channel. So to have internnaly double=True and timeaxis=1. This garanty no copy at reading but at a cost of double writting and transposing.

Could I make a proposal starting from your branch ?

campagnola: I have been thinking a lot about this. Let me see if I understand: The default is to have timeaxis=0 (sample, channel), which is trivial to write into a ring buffer. Right now using timeaxis=1 (channel, sample) does not make sense in RingBuffer because the data does not stack nicely along axis 0. In this case, are you suggesting splitting up the chunk as it is written into the buffer, such that each channel is written into a different part of the buffer to ensure they remain contiguous? For example, let's say the data is 4 samples by 2 channels:

    [1, 1, 1, 1, 2, 2, 2, 2]

If we write this into a ring buffer of length 8, then it should look like:
 
    [1, 1, 1, 1, 0, 0, 0, 0, 2, 2, 2, 2, 0, 0, 0, 0]

Is that correct? 
Why didn't I understand this sooner? 
I might not need the 1D ring buffer at all...

samuelgarcia: Yes this is what I had in mind.
And yes I think the ringbuffer is just a special case a NDRingbuffer.

For me all the code go in the good way in RingBuffer, we just need to extend ND case in RingBuffer by playing on the dimenssion 0 in new_chunk and get_data.

But overall, we need  we need a swapaxes kargs which is None by default. (timeaxis=0)

In the case of (the case i want) swapaxes is (0,1) (timeaxis=1 internally), just after creating the np.array we perform aswapaxes in the __init__ so that when we have a new chunk working on the dimension 0, it is silently done on timeaxis 1 by numpy magic index and strides.



And the goal is this one:
Case continous by sample:
```
length = 1024
nb_channel = 16
ringbuf = NDRingBuffer(shape=(length, nb_channel), double=True, swapaxis=None)
# internally it is
self.buf = np.ones(nb_channel, length)

# so that when

chunk = ringbuf.get_data(start, stop)
chunk[:, chan] # this is not continuous in memory!!!!!!
chunk[pos, :] # this is continuous in memory
```



Case continous by channel:
```
length = 1024
nb_channel = 16
ringbuf = NDRingBuffer(shape=(length, nb_channel), double=True, swapaxis=(0,1))
# internally it is
self.buf = np.ones(nb_channel, length)  # note that there is little hack on the shape which is also swaped
self.buf = self.buf.swapaxis(0,1)

# so that when

chunk = ringbuf.get_data(start, stop)
chunk[:, chan] # this is continuous in memory!!!!!!
chunk[pos, :] # this is not continuous in memory
```

Really sorry to not be clear. I guess this is partly due to my english but you know my level...

campagnola: Not at all! I think you explained it clearly many times, but I kept seeing this:

    [1111, 2222] + [1111, 2222] = [1111, 2222, 1111, 2222]

(which doesn't make any sense) instead of this:
 
    [1111, 2222] + [1111, 2222] = [11111111, 22222222]
 
I might revert back to the point where I had a fully functional ND ring buffer. And I finally almost agree that we need `swapaxes`  :D

samuelgarcia: Oh yeah.
Do you need help ? Or prefer finish that branch first ?

campagnola: I would not mind some help, but I think I am going to revert this branch back to an earlier state where it will be easier to implement swapaxes, so I don't recommend making any changes directly from here.. I'll let you know when I find a good starting place.

campagnola:
Update: I reverted back several commits to a simpler version of RingBuffer. From there it was almost trivial to set the memory order of the buffer.

I came across an issue with `swapaxes`: if you have a 3D stream (maybe a 2d electrode array or a video stream) and you want the channels to be contiguous in memory, then you would need `swapaxes=0,2` to make the time axis last. But perhaps you want to keep the order of the (1,2) axes intact. In that case, you need to be able to specify the order of all three axes (1, 2, 0). So instead of `swapaxes`, I added an `axisorder` option. It works like:

* For plaindata streams, data is sent in exactly its original memory order because this is the most efficient. If the receiver has a ring buffer attached, though, then the memory layout in the buffer will be determined by `axisorder`.
* For sharedmem streams, the shared ring buffer uses `axisorder` for its memory layout.
* Both stream types have the same API for accessing data chunks as they arrive, or via ring buffer.

Let me know what you think..

samuelgarcia:
This is excactly what I wanted but I was to shy for asking.
Gogogo, you are the best.

campagnola:
You asked me for this about 6 times.  I just thought I could make it work without swapaxes until I finally understood the ringbuffer solution  :)

samuelgarcia:
I was mentioning the axisorder . :)

pyacq/core/stream.py
```diff
@@ -505,29 +572,400 @@ def __init__(self, socket, params):
...
+        """
+        assert size <= self.size, ("Requested data size (%d) is larger than ring buffer (%d)." %
+                                   size, self.size)
+        if self.double:
+            ind = (self._ptr - size) % self.size
+            data = self.buffer[ind:ind+size]
+            if copy:
+                data = data.copy()
+        else:
+            ind = self._ptr - size
```
samuelgarcia: Proposal: in the case double=False and ind<0 to avoid copy.
We could add an optional always_split=False.
get_data(self, size, copy=False, always_split=False):

always_split=False: is the case implemnted here
always_split=True return always 2 parts :
data1, data2 = RingBuffer.get_data(...)
where data2 is None except when ind<0.

pyacq/core/stream.py
```diff
@@ -285,84 +284,157 @@ class PlainDataSender:
...
+                
+        # serialize
+        dtype = data.dtype
+        shape = data.shape
+        buf, offset, strides = decompose_array(data)
+        
+        # compress
+        comp = self.params['compression']
+        if comp.startswith('blosc-'):
+            assert HAVE_BLOSC, "Cannot use blosclz4 compression; blosc package is not importable."
```
samuelgarcia: This is buggy on my machine.
blosc need bytes for compression.
Could you also fix this in the PR ?

pyacq/core/stream.py
```diff
@@ -285,84 +284,157 @@ class PlainDataSender:
...
+        stat = struct.pack('!' + 'Q' * (3+len(shape)) + 'q' * len(strides), len(shape), index, offset, *(shape + strides))
+        copy = self.params.get('copy', False)
+        self.socket.send_multipart([stat, buf], copy=copy)
     
     def close(self):
         pass
 
 
-class PlainDataReceiver:
+class DataReceiver:
```
samuelgarcia: Could we also add the kargs ring_axisorder=None.
The use case could be that a Node wants a diffrents layout in memory than the sender Node for instance for GPU computing...
So if:
  * ring_axisorder=None, params['axisorder'] is takken
  * ring_axisorder=(...) the newaxis is use in the ringbuffer proxy.

pyacq/core/stream.py
```diff
@@ -285,84 +284,157 @@ class PlainDataSender:
...
+        if self._ring_buffer is not None:
+            self._ring_buffer.new_chunk(chunk, index)
+        return index, chunk
+
+    def __getitem__(self, *args):
+        if self._ring_buffer is None:
+            raise TypeError("DataReceiver indexing requires a ring buffer.")
+        return self._ring_buffer[args]
+    
+    def get_array_slice(self, index, length):
```
samuelgarcia: Maybe we could remove this no ?

pyacq/core/stream.py
```diff
@@ -285,84 +284,157 @@ class PlainDataSender:
...
     
-    def _uncompress_blosc(self, index, data):
-        data = blosc.unpack_array(data)
-        return index, data
+    def __getitem__(self, *args):
+        # make shm buffer publicly readable
+        return self._buffer[args]
 
 
 class SharedArraySender:
```
samuelgarcia: I guess we can remove SharedArraySender/SharedArrayRceveiver
No ?

pyacq/core/stream.py
```diff
@@ -285,84 +284,157 @@ class PlainDataSender:
...
+        assert data.dtype == self.params['dtype']
+        shape = data.shape
+        if self.params['shape'][0] != -1:
+            assert shape == self.params['shape']
+        else:
+            assert shape[1:] == self.params['shape'][1:]
+ 
+        self._buffer.new_chunk(data, index)
+        
+        stat = struct.pack('!' + 'Q' * (2+len(shape)), len(shape), index, *shape)
```
samuelgarcia: Why sending all theses and not only the index ?

pyacq/core/stream.py
```diff
@@ -285,84 +284,157 @@ class PlainDataSender:
...
+    def __init__(self, socket, params):
+        # init data receiver with no ring buffer; we will implement our own from shm.
+        DataReceiver.__init__(self, socket, params)
+
+        self.size = self.params['buffer_size']
+        shape = (self.size,) + self.params['shape'][1:]
+        self._buffer = RingBuffer(shape=shape, dtype=self.params['dtype'],
+                                  shmem=self.params['shm_id'], axisorder=self.params['axisorder'])
+
+    def _recv(self):
```
samuelgarcia: In the old behavior for sharedarrayreceiver, the data was always None and only the index had sens.

The reasons was : in sharredarray buffer, often the last chunk have no meaning computations are done on longer chunk sliding chunk. So we need a case where data is not constructed at each send/recv. I guess have no proof that intensive send/recv this may count to not construct last data chunk.
So I propose a kargs in init: with_data=True/False and with False 
```data = self._buffer[index+1-shape[0]:index+1]``` would be skipped.

**Comment by samuelgarcia** [comment]
I feel happy with that changes.
Could you also split in two files (stream and ring) as we discuss it previously ?
We are near something good no ?

File: pyacq/core/stream.py (hunk 5 above)
Comment (campagnola) [inline pyacq/core/stream.py]: I think this is taken care of now.

**Comment by campagnola** [comment]
Split classes into a new stream subpackage. Still have several of your comments to address, and nodes need to be adapted (want to help? See changes to qoscilloscope--mostly just deleted code and a few minor changes).

This is getting close!

File: pyacq/core/stream.py (hunk 5 above)
Comment (samuelgarcia) [inline pyacq/core/stream.py]: I think no.
This line make a new array:
data = np.empty((size,) + self.buffer.shape[1:], dtype=self.buffer.dtype)
my idea is:
```
    if self.double:
        same as previous
    else:
        if always_split:
            if ind < 0:
                data1 = self.buffer[ind:]
                data2 = self.buffer[:self._ptr]
            else:
                data1 = self.buffer[ind:ind+size]
                data2 = None
            return data1, data2
        else:
           #same as previous

samuelgarcia:
Yes I want to help.
I can take care all viewer/dsp Nodes with this new design.
I will do PR directly on your place and so we will have one only PR for this new streamm design.
OK ?

pyacq/core/stream.py
campagnola: This is out of date; check the latest: core/stream/ringbuffer.py

pyacq/core/stream.py
campagnola: It works for me, so I'm not sure what to change? Something like 

    buf = blosc.compress(buf.tostring(), data.itemsize, cname=comp[6:])

?

pyacq/core/stream.py
campagnola: Right now this is handled when you call `InputStream.set_buffer(..., axisorder=...)`. I thought about making this configurable through `InputStream.__init__()`, but then we would need `ring_axisorder`, `ring_double`, and `ring_size`. What do you think?

campagnola: Aside: I was also thing about making other types of buffer, which in part led me to think that each buffer should take care of its own arguments.

pyacq/core/stream.py
campagnola: I was going to ask.. 
Done.

pyacq/core/stream.py
campagnola: Done.

pyacq/core/stream.py
campagnola: Done.

pyacq/core/stream.py
campagnola: Done. What do you think of `return_data` instead? I thought `with_data` was a little confusing.

campagnola:
Yeah, go ahead and start a sub-PR from this one. I won't rebase anymore.
Meanwhile, there are tests for me to fix..

pyacq/core/stream.py
samuelgarcia: yep.
Sorry

pyacq/core/stream.py
samuelgarcia: I reinstall blosc and everything is OK now.
Sorry.

pyacq/core/stream.py
samuelgarcia: I have same idea.
I would love a Sender/Revever class:
   1- opencl (or cuda) buffer
   2- local ringbuffer but with no socket just with callback. Send a packet directly call recv.

We can postpone for later no ?
Your new design is really clean and let space for lot ideas.

pyacq/core/stream.py
samuelgarcia: OK for return_data.

samuelgarcia:
I am super happy. This new design really smart. Thanks a lot.
I have only one small conern about file names: 
pyacq/core/stream/sharedarray.py and pyacq/core/stream/sharedmem.py are similar names BUT are differents beast.
I propose :
```
git mv yacq/core/stream/sharedmem.py yacq/core/stream/sharedmemstream.py
git mv yacq/core/stream/plaindata.py yacq/core/stream/plaindatastream.py
```

pyacq/core/stream/ringbuffer.py
```diff
@@ -0,0 +1,304 @@
...
+                    data[break_index-start:] = b
+                    copied = True
+        
+        if copy and not copied:
+            data = data.copy()
+            
+        if join:
+            return data
+        else:
+            empty = np.empty((0,) + data.shape[1:], dtype=data.dtype)
```
samuelgarcia: Maybe ```empty=None``` could be enougth here no ?

pyacq/core/stream/ringbuffer.py
```diff
@@ -0,0 +1,304 @@
...
+            raise IndexError("Requested segment (%d, %d) is out of bounds for ring buffer. "
+                             "Current bounds are (%d, %d)." % (start, stop, first, last))
+        
+        bsize = self.shape[0]
+        copied = False
+        
+        if self.double:
+            start_ind = start % bsize
+            stop_ind = start_ind + (stop - start)
+            data = self.buffer[start_ind:stop_ind]
```
samuelgarcia: why not a ```return data``` here.
This would avoid join mechanism in doule=True case.
no ?

samuelgarcia:
I have a question for InputStream.set_buffer in case of SharedMemReceiver.set_buffer.
If I understand correctly, this create a new RingBuffer that is not the original sharedmemory so a new copy is done at each pacquet. No ?

Is it what we want. Won't it be safer to put set_buffer only for PlainDataReceiver.set_buffer ?

pyacq/core/stream/stream.py
```diff
@@ -0,0 +1,294 @@
...
+    def __getitem__(self, *args):
+        """Return a data slice from the RingBuffer attached to this InputStream.
+        
+        If no RingBuffer is attached, raise an exception. See ``set_buffer()``.
+        """
+        if self.buffer is None:
+            raise TypeError("No ring buffer configured for this InputStream.")
+        return self.buffer.__getitem__(*args)
+
+    def set_buffer(self, size=None, double=True, axisorder=None):
```
samuelgarcia: To be sure of the wanted behavior here:
  * If transfermode=='plaindata' : I create a local RingBuffer just the Node that use this input.
  * If transmode=='sharedmem': the buffer of receiver (so the shared memory) is taken if the size is enougth and axisorder iscompatible. If not create a local one.

Is that correct ?

campagnola: Isn't that the existing behavior? When you call `set_buffer`, the first thing it does is to check whether the inputstream or the receiver already has a buffer that meets the requested parameters.

However, we could have it automatically set the inputstream buffer so that it is not necessary to call set_buffer with sharedmem..

samuelgarcia:
I am back.

In ```class SharedMemReceiver(DataReceiver):```

This ```data = self.buffer[index+1-size:index+1]``` make me worry.
Why is that +1 for.

Are we agree that when receive ```index```, ```index``` is not included this is the upper bound like any python spirit buffer manipulation.

In short, if index is 10, this means that I have 10 samples in the buffer and I when to slice ```buffer[0:10]``` so why this +1 for ?

pyacq/core/stream/stream.py
samuelgarcia: Yes making it automatic could be good no ?

pyacq/core/stream/stream.py
```diff
@@ -0,0 +1,294 @@
...
+            self.node().after_output_configure(self.name)
+
+    def send(self, data, index=None, **kargs):
+        """Send a data chunk and its frame index.
+        
+        Parameters
+        ----------
+        index: int
+            The absolute sample index. If the chunk contains multiple samples,
+            then this is the index of the last sample.
```
samuelgarcia: Before this PR, index was the index of last sample+1, in short the upper bound excluded.
I think I prefer before...
Sorry to be so late for commenting that.

This is more intuitive for a python coder. The index was NOT (like here) the included upper bound but index is absolute length (so the excluded upper bound).

Do you known that we could banished from the python community with such a choice ?

What do you think.

samuelgarcia:
I have a small problem with the new design.

I try to explain.

On a viewer I want the last 1000 samples but my index_head=100.
So I expect to have 900 samples filled with nan or 0 (depending filler=..) and 100 samples with the valid samples.

I can do ```self.input[-10000:]``` and this works. This is done in oscilloscope.

I would also like to write ```self.input[index_head-10000:index_head]``` because sometimes ```index_head=index_head%downsampling_factor```.
This case do not work because index_head-10000 is negative.
This come from ```RingBuffer._interpret_index``` do you think we could fix that easily ?

Before this was handle by playing with the stop and not the start.

Something like (see SharedArrayReceiver.get_array_slice in master)

```
if double and stop>0:
    length=stop-start
    i2 = stop%self.shape[0] + self.shape[0]
    i1 = i2 - lenght
   buffer[i1:i2]
```

samuelgarcia:
I have another problem with the new design.

In case of transfermode='sharedmem' if the receiver do a ```self.input.set_buffer(..)``` like in oscilloscope.
Then this add an extra complexity to ensure that there is extra work to recreate a each packet an array that already exists in the poller

See new changes in Oscilloscope.

pyacq/core/stream/stream.py
campagnola: HMMMM. Maybe not; I think this actually makes the oscilloscope code more complex.

In general, the oscilloscope should not know or care what kind of stream it is connected to; it always needs to ensure that a ring buffer of sufficient size is availaable. If it is connected to plaindata, then a new ring buffer needs to be created. If it is connected to sharedmem but the shared buffer is too small, then again a new buffer needs to be created. 

So rather than have the oscilloscope (and every other input node) check to see whether a large-enough buffer is available, we just have it declare, "I need a ring buffer with these properties" and let the inputstream decide whether to create a new one.

pyacq/core/stream/stream.py
campagnola: Excellent!! I thought I was just preserving the old API, and it was really awkward. I should have just asked about the docstring "this is the index of the last sample" when I first saw it. I will gladly fix this, so neither of us has to endure exile from python.

However: what do you think about making this the *first* index of the chunk? That way:

    videostream.send(first_frame, index=0)  # instead of index=1

campagnola:
> I would also like to write self.input[index_head-10000:index_head] because sometimes index_head=index_head - index_head%downsampling_factor for alignement case when downsampling. This case do not work because index_head-10000 is negative.

I think we should keep the negative index behavior as it is, because this is consistent with how indexing usually works in arrays/lists. If you want to be able to request a truly negative index, then we should expose `RingBuffer.get_data()` instead.

> In case of transfermode='sharedmem' if the receiver do a self.input.set_buffer(..) like in oscilloscope.
Then this add an extra complexity to ensure that there is extra work to recreate a each packet an array that already exists in the poller

I'm not sure I understand (because I thought I answered this already), but .. `InputStream._own_buffer` is a flag that prevents the stream from copying data into the buffer. So if we have a sharedmem receiver, and the receiver's buffer is sufficiently large, then the inputstream will just use that existing buffer and avoid making any unnecessary copies (at least, that is how it is supposed to work; perhaps there is a bug). Is this the issue you are worried about, or did I misunderstand?

pyacq/core/stream/stream.py
samuelgarcia: Sorry for the docstring that was confusing.

I guess that +1 in RingBuffer should also disappear.

I prefer the last index not included. So the index is also the total length.
It sound more natural. In general for DAQ api, the sementic of 'index' or 'head' are often upper not included bound.

samuelgarcia:
The mecanism with ```InputStream._own_buffer``` is OK.
I try to explain better. I am confused too.
Have a look to my sub PR.

I try step what make me incofortable:
  1 - I did not like the the core.tool.ThreadPollInput was always receving data even for shermem. Bad performences, a np.array is constructed at each packet.
  2 - I forced SharedMemReceiver.recv(self, return_data=False) by default. This is more natural. We do not need to reconstruct the last chunk in that case event if it is a view this could lead overhead. We often need a larger buffer less but less otfen. I hope you will be ok with that.
  3 - So I needed to introduce ```core.tools.ThreadPollInput.__init__(..., with_data=True/False, ...)```
  4 - So I needed to indroduced also ```with_data``` in PlainDataReceiver.recv(with_data=True). this could make sens in some case and we just need the index and not the data. This was mostly to have same API for the 2 streams class.
  5 - I needed to take care of with_data=True/False in ThreadPollInput depending the case. If False this make that input.buffer is not filled in plaindata so we need True. If False this make construct the last chunk for sharedmem this is have performence impact so we need False.
  6 I finally handle this Oscillscope wich is ungly:
```
        # poller
        if  self.input.params['transfermode']=='sharedmem':
            return_data=False
        else:
            return_data=True
        self.poller = ThreadPollInput(input_stream=self.input, return_data=return_data)

```

**Conclusion:**

Writting this make me stupid. We could deal with directly in ThreadPollInput with

```
ThreadPollInput.__init__(.., with_data=None, ...)
```
None means  'auto' and the ugly things is done at pooler level.

It was not obvious last nigth.
I try it and get back.
At the end everything is OK.

pyacq/core/stream/ringbuffer.py
```diff
@@ -0,0 +1,304 @@
...
+            start, stop, step = self._interpret_index(first)
+            data = self.get_data(start, stop)[::step]
+            if rest is not None:
+                data = data[rest]
+        else:
+            raise TypeError("Invalid index type %s" % type(first))
+        
+        return data
+
+    def get_data(self, start, stop, copy=False, join=True):
```
samuelgarcia: I more often need get_data from stop and  lentgh.
This was is ex-behavior with_get_array_slice.
So I very often do this ```get_data(stop-length, lentgh)```.
I have no opinion on that but my intuition lead to prefer have ```get_data(stop, length)```  directly. Even if it sounds more natural to have ```get_data-stop, start)```

pyacq/core/stream/stream.py
campagnola: Fixed! Surprisingly easy, and I got to delete SO MANY "`+ 1`".

samuelgarcia:
I have finish the sub PR on your account
https://github.com/campagnola/pyacq/pull/4

There are still some monior problem arround InputStream.connect when checking if params correspond to spec. For shape this is bugguy for some pair of device<->viewer because the device have shape (1024, 2) and the viewer have (1, 2). This occours for audio_pyaudio, webcam_av and certrainly more. Have a look you will understand. I have no idea the best way to fix. ignore when -1 in shape. In that case shape would be a special case of readonly_params. 

Otherwise, have a look in dsp/* and device/*. This is the main work from me here.
If you are happy you can merge into your big PR.

Them we should be OK for merging in the master.
And so come back on many other fronts.

Thanks a lot.

samuelgarcia:
This PR is replace by PR #51.

PR #51, opened 2017-04-19
Stream performance with samuelgarcia commits

samuelgarcia:


samuelgarcia:
I force the PR even if travis is not working. This is due to other things in the master.
This will be fixed later on.

PR #52, opened 2017-04-25
There is a bug for atexit

samuelgarcia:
There an issue with atexit because some atexit are current:
  * the one in ProcessSpawner
  * the one in Host
  * the one in create_manager
  * the one in Manager.__init__

This must be fixed for renning all tests.


samuelgarcia:
This branch is now useless.

PR #53, opened 2017-04-29
Deadlock bugfixes

campagnola:
Corrections for two deadlocks that could happen during exit. 
This seems to fix all of my current `atexit` issues.


campagnola:
Hmm.. maybe spoke too soon.

campagnola:
Looking better. 
@samuelgarcia, does this resolve atexit issues for you?

samuelgarcia:
This truly improve the situation and this is so good!!

Thank you.

PR #41, opened 2015-12-02
Appveyor CI (2)

campagnola:
Work in progress..

PR #55, opened 2017-05-02
Windows bugfixes

campagnola:
WIP: getting tests passing on windows.
Depends on #53.

### DIFF
--- a/pyacq/core/rpc/client.py
+++ b/pyacq/core/rpc/client.py
@@ -1,3 +1,4 @@
+import sys
 import os
 import time
 import weakref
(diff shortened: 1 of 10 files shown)


campagnola:
@samuelgarcia, I replaced all instances of `tcp://*:*` with `tcp://127.0.0.1:*`. The main reason is that using a wildcard for the address causes the `socket.last_endpoint` to look like `0.0.0.0`, which causes the client to barf in windows (linux handles this with no trouble). An additional benefit is that pyacq servers are not automatically opening up to the internet by default. 

The downside is that if you _wanted_ to have communication across machines on your network, then you now have to manually specify the IP address to bind to (or `*:*` on linux).

samuelgarcia:
Thank you for that fix on windows.

Is the old PR #41 still is valid ?

samuelgarcia:
This is so good!!!

PR #61, opened 2017-05-06
Make client check for local thread server on every read.

campagnola:
Otherwise, a client created before the server in the same thread will not
allow reentrancy.


samuelgarcia:
Hi Luke.

Thank you for continuing efforts.

Can I merge this ?

campagnola:
Yes, this one is ready.

PR #63, opened 2017-07-28
Make pyaudio_oscope_local.py example repeatable

fjaragones:
# pyaudio_oscope_local.py - main: stop and close pyaudio node to properly close pyaudio
# Otherwise, an error appears in succesive executions: 
# ValueError: Input not supported: channels=1 samplerate=44100.0 device_id=2

# audio_pyaudio.py - PyAudio._close
# Condition added to address the fact that thread is only created when self.output_device_index is not None
# otherwise: AttributeError: thread

# stream.py - InputStream.close
# There are situations in which input does not need to be connected (namely audio_pyaudio node) and hence, trying to access for closing produces an error
#  AttributeError: 'InputStream' object has no attribute 'receiver'
#  Try...catch added around- self.receiver.close()
# same for socket
# Evaluate if the same applies to OutputStream.close

### DIFF
--- a/examples/pyaudio_oscope_local.py
+++ b/examples/pyaudio_oscope_local.py
@@ -54,3 +54,7 @@ if __name__ == '__main__':
     import sys
     if sys.flags.interactive == 0:
         app.exec_()
+
+# stop and close pyaudio node to properly close pyaudio
+dev.stop()
+dev.close()
--- a/pyacq/core/stream/stream.py
+++ b/pyacq/core/stream/stream.py
@@ -297,9 +297,12 @@ class InputStream(object):
         This closes the socket.
         
         """
-        self.receiver.close()
-        self.socket.close()
-        del self.socket
+        try:
+            self.receiver.close()
+            self.socket.close()
+            del self.socket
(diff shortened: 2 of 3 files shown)


examples/pyaudio_oscope_local.py
```diff
@@ -54,3 +54,7 @@
     import sys
     if sys.flags.interactive == 0:
         app.exec_()
+
+# stop and close pyaudio node to properly close pyaudio
```
campagnola: If this is necesaary, then perhaps the device should handle it with an `atexit` callback?

fjaragones: I've tried registering cleanup code in Pyaudio-__init__
But I'm launching the example from Spyder IDE and it looks like cleanup code only gets called when you restart the kernel. 
There are also issues whith using instance methods as cleanup code: https://stackoverflow.com/questions/16333054/what-are-the-implications-of-registering-an-instance-method-with-atexit-in-pytho
Anyway, the stop and close way should be an option and not produce errors (the ones that the other proposed changes fix)

PR #69, opened 2017-10-12
Stream docs

campagnola:
Cleaned up / expanded docstrings for stream classes.

### DIFF
--- a/pyacq/core/stream/sharedmemstream.py
+++ b/pyacq/core/stream/sharedmemstream.py
@@ -17,7 +17,6 @@ class SharedMemSender(DataSender):
     
     Extra parameters accepted when configuring the output stream:
     
-    * shape (tuple) The shape of one frame (a sample at one time point).
     * buffer_size (int) the size of the shared memory buffer in *frames*.
       The total shape of the allocated buffer is ``(buffer_size,) + shape``.
     * double (bool) if True, then the buffer size is doubled and all frames are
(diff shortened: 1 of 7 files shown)


samuelgarcia:
Hi luke.
Good to see you back here.

Thank you for doc.

Tell me when ready to merge.

I will add some comment in review soon.

pyacq/core/stream/sharedmemstream.py
```diff
@@ -9,7 +9,25 @@
...
-    """
+    """Stream sender that uses shared memory for efficient interprocess
+    communication. Only the data pointer is sent over the socket.
+    
+    Note: this class is usually not instantiated directly; use
+    ``OutputStream.configure(transfermode='sharedmem')``.
+    
+    Extra parameters accepted when configuring the output stream:
+    
+    * shape (tuple) The shape of one frame (a sample at one time point).
```
samuelgarcia: For multi channel, we can put that shape=(-1, nb_channel) is a good practice.

This avoid to mix up with buffer_size.

campagnola: Ah, good catch. I removed shape from the sender-specific docs, since the OutputStream docs already cover it correctly.

campagnola:
It looks ready to me.. did I miss anything?

samuelgarcia:
No this is perfect!! You are the boss!!

PR #70, opened 2017-10-12
setup.py automatically finds the packages it should install

campagnola:


samuelgarcia:
I guess this depend on #71 because there are 2 commits.

campagnola:
No relation to #71. I think this is ready to merge.

PR #71, opened 2017-10-12
WIP: logging / debugging tools

campagnola:
Adds  ui for viewing log message streams from multiple hosts. This is intended to make it easier to understand errors that occur during distributed operation.

Some issues still need to be worked out:
 - best way to handle exceptions in remote processes
 - best way to handle text wrapping in the log viewer


samuelgarcia:
@campagnola : Can we merge this  or do you want improve more ?

campagnola:
I think this is ready to go; further improvements can come later.

PR #72, opened 2017-12-07
Nidaqmx

samuelgarcia:
New Nidaqmx Node based on the official python wrapepr from ni.

### DIFF
--- a/examples/ni_daqmx_with_oscilloscope.py
+++ b/examples/ni_daqmx_with_oscilloscope.py
@@ -22,7 +22,7 @@ aochannels =  ['Dev1/ao0', 'Dev1/ao1']
 
 #create 2 output short signals than will be played every seconds
 times =np.arange(0,0.5, 1./sr)
-freq, ampl = 1.5, 2.5
+freq, ampl = 5., 2.5
 sig0 = np.sin(2*np.pi*freq*times).astype('float64')*ampl
 sig1 = np.random.randn(sig0.size).astype('float64')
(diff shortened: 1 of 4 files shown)


campagnola:
Looks like there is some code to support analog output channels.. is this just not finished yet?

samuelgarcia:
For the first round it is finished. Until the enxt bug.

The scenario is: coninuous acquisition on input. And episodic play some signal on output (at the same sample rate). So that feed my needs for the moments.

We should also add digital input/output but this will be for next time.

campagnola:
Error running the example:

```
  File "c:\users\lukec\pyacq\pyacq\devices\ni_daqmx.py", line 232, in run
    nb_sample = aitask.in_stream.readinto(raw_data_flat)
  File "C:\Anaconda3\lib\site-packages\nidaqmx\_task_modules\in_stream.py", line 1992, in readinto
    self.timeout)
  File "C:\Anaconda3\lib\site-packages\nidaqmx\_task_modules\read_functions.py", line 583, in _read_raw
    ctypes.byref(number_of_bytes_per_sample), None)
ctypes.ArgumentError: argument 2: <class 'TypeError'>: wrong type
```

samuelgarcia:
You need this: https://github.com/ni/nidaqmx-python/pull/45

campagnola:
Thanks! Works well on my end with a USB-6259.

samuelgarcia:
I will merge this.

PR #73, opened 2017-12-14
SImple node recorder: RawRecorder.

samuelgarcia:


campagnola:
Looks good to me, although it would be really nice to have a class (or functions) that formalize this new data format. It'd be ok to have that in a separate PR, though.

samuelgarcia:
Formating data need some clarification at stream level with gain/offset/dtype be able to store in raw format (int16) or in scaled format (float32/64) with units.

What is the choice for that in acq4 ?

For next time, I think.

campagnola:
ACQ4 stores data as HDF5 so shape/dtype are not issues, and all scale/offset are applied immediately after acquiring the data. It would be nice here just to have a function that can correctly recreate a memmapped array from the stored structure.

samuelgarcia:
Iw ill merge.

PR #74, opened 2017-12-15
Bug with complex dtype with msgpack.

samuelgarcia:


samuelgarcia:
I need this merge very soon for a demo on Monday. So I do it it now...

PR #80, opened 2018-03-12
Multi channel system W2100

samuelgarcia:
Implement a Node for Multi Channel System W2100 system


campagnola:
Yeah, go ahead and take out faulthandler. You only need it if you expect to be debugging segmentation faults. 

I'm curious, though: could you just call faulthandler.disable() before importing clr? That way it would always be enabled except when pythonnet is used.

PR #87, opened 2018-06-06
Blackrock

samuelgarcia:
This is a new Node to grab continuous from blackrock device Neural Signal Processing (NSP).
This is based on there DDL with ctypes (and some hacks).
Event and spike are NOT grabbed from the moment.


samuelgarcia:
I plan to merge this very soon.
If someone want to make some coment don't wait too much :)

PR #91, opened 2018-06-20
patch to pass devices unit testing

Lx37:
simple pytest skip to avoid runing unit tests on device tests.
Maybe a general flag could be used to know if the tests runs from travis/appveyor, to skip device testing properly? 

something inspired by
@pytest.mark.skipIf("TRAVIS" in os.environ and os.environ["TRAVIS"] == "true", "Skipping this test on Travis CI.")

### DIFF
--- a/pyacq/devices/tests/test_eeg_openBCI.py
+++ b/pyacq/devices/tests/test_eeg_openBCI.py
@@ -12,6 +12,7 @@ from pyqtgraph.Qt import QtCore, QtGui
 
 import pytest
 
+@pytest.mark.skip(reason="need Device")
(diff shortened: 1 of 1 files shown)


campagnola:
I wonder two things:
* Is it possible to check to see if the device is available, and if not, skip the test?
* Is it safe to allow unit tests to access acquisition hardware by default? I wonder if all hardware tests should require an explicit opt-in. like `PYACQ_TEST_HARDWARE=1 py.test ...`

samuelgarcia:
I agree to add a global varaible PYACQ_TEST_HARDWARE that would be False on travis and appveyor.
I merge this and add an issue.

PR #89, opened 2018-06-07
Rt spike expe

samuelgarcia:
This branch need to be merged after:
  * #87 
  * #86 
  * #88

PR #94, opened 2018-06-27
Try debug on window and appveyor.

samuelgarcia:


samuelgarcia:
Ni Luke,
I am trying to debug appveyor problem in a separate PR.
We have problems in:
  * test_rpc.pt test_disconnect()
  * test_npbufferdevice.

I think problems are linked to atexit in ProcessSpawner that try to stop something that do no exists anymore and so the socket is aleady dead.
Do you have any idea ?

We alaso have problem in:
  * test_sosfilter
  * test_analogtrigger
I think here it is another problem.

Best,

samuelgarcia:
Problem in appveyor seams to be random...
They change at each build...
Bad news....

Now there problem in QTimerfreq. Some of them are already fixed in #89.

samuelgarcia:
This pass test.
I am lucky.
I merge.

PR #95, opened 2018-06-27
Update travis version of miniconda and python

samuelgarcia:


samuelgarcia:
Need to be merge before #89

